PR #2, opened 2016-11-19
Fix my_rooms -> myRooms

OverlordQ:
fix reference to myRooms


TooAngel:
Oh, thank you for the PR I really appreciate it. Sorry, I somehow missed it and it should be fixed in my last commit.

PR #39, opened 2016-12-26
Fix #27 and add local config file

### LINKED ISSUE #27: Scout stuck on border of closed rooms
If a room is closed, scouts will stay on borders and because the room will never be registered, all next scouts will do the same.

WhiteHalmos:


WhiteHalmos:
Closed in favour of https://github.com/TooAngel/screeps/pull/40

PR #40, opened 2016-12-26
Stuff instead of master

WhiteHalmos:
Maybe this is better as you'll be merging in stuff anyways.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .env
 node_modules/
 dist/
-config_local.js
--- a/README.md
+++ b/README.md
@@ -35,10 +35,6 @@ while fighting or other occasions which needed quick fixes or in the ingame
 editor. But I think there are a couple of funny ideas. Every contribution is
 welcome.
 
-Copy `src/config_local.js.example` to `src/config_local.js` and use it to
-modify configs for your own bot rather than directly modifying `src/config*.js`
-defaults.
-
 ## Features
 
  - Automatic base building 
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 .env
 node_modules/
 dist/
+config_local.js
```
TooAngel: Not sure why you remove it in the room `.gitignore`.

I think I would either add `src/config_local.js` to this file, or create a `.gitignore` in the `src/` directory with `config_local.js`

README.md
```diff
@@ -33,7 +33,11 @@ market.
 This is not a good example for code quality or structure, many LOC are written
 while fighting or other occasions which needed quick fixes or in the ingame
 editor. But I think there are a couple of funny ideas. Every contribution is
-welcome. 
+welcome.
+
+Copy `src/config_local.js.example` to `src/config_local.js` and use it to
```
TooAngel: We could put the `config_local.js.example` in the root. So that in the `src/` directory are only `.js` files. No strong opinion on this, just an idea.

src/config.js
```diff
@@ -14,7 +14,7 @@ try {
 
 global.config = {
   profiler: {
-    enabled: true
+    enabled: true,
```
TooAngel: I like the `config_local.js` idea :+1: 

I think the default values should be the bot values. No profiler, no summary, no log, no nothing. So we can change that to false.

src/require.js
```diff
@@ -3,6 +3,7 @@ require('logging');
 require('config_brain_memory');
 require('config_brain_nextroom');
 require('config_brain_squadmanager');
+require('config_local');
```
TooAngel: This needs to be encapsulated in a `try{ } catch (e) {}` block, if no `config_local.js` exit.

.gitignore
WhiteHalmos: Done.

README.md
WhiteHalmos: Done

src/require.js
WhiteHalmos: Done, moved to config.js like friend.js since other config files are methods.

src/config.js
WhiteHalmos: Done

WhiteHalmos:
Updated, PTAL

Issue #36, opened 2016-12-23
Improve code structure


TooAngel:
Many for loops can be replace by lodash filter.

Checks for structures can be replaced by tables:
```
     let hostileStructures = creep.room.find(FIND_HOSTILE_STRUCTURES, {
        filter: function(object) {
          let table = {
            [STRUCTURE_RAMPART]: true,
            [STRUCTURE_EXTRACTOR]: true,
            [STRUCTURE_WALL]: true,
            [STRUCTURE_CONTROLLER]: true
          };
          return !table[object.structureType];
        }
      });
```

WhiteHalmos:
I like these tables, quite clear.

Nevrdid:
thinking about better structure ... 
`
let hostileStructures = creep.room.find(FIND_HOSTILE_STRUCTURES, { 
  filter: struct=> _.include([STRUCTURE_RAMPART,
                                      STRUCTURE_EXTRACTOR,
                                      STRUCTURE_WALL,
                                      STRUCTURE_CONTROLLER], struct.structureType)
});
`

mschultheiss83:
and go a bit further by move these arrays to
```roles.[role].HOSTILE_STRUCTURES``` or to 
```global.config.HOSTILE_STRUCTURES```

Nevrdid:
yep, i can add it with roles.[role].settings. will check that.

TooAngel:
I started with something like this some time ago: https://github.com/TooAngel/screeps/blob/master/src/prototype_room_cached.js

I don't think it is good to bind it to the room object, better to store in a global variable. In my cases I wanted it refreshed on each tick. I think we need to check how often it needs to be refreshed.

TooAngel:
I think we have some kind of solution with the find abstractions.

PR #44, opened 2016-12-27
Fix harvester logic

WhiteHalmos:
Better basic harvesters so only need 2-4 rather than 5 or more.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -32,8 +32,8 @@ global.config = {
   },
 
   autoattack: {
-    notify: false,
     disabled: false,
+    notify: false,
   },
 
   nextRoom: {
--- a/src/config.js
+++ b/src/config.js
@@ -32,7 +32,8 @@ global.config = {
   },
 
   autoattack: {
-    disabled: false
+    notify: false,
+    disabled: false,
   },
 
   nextRoom: {
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -702,7 +702,7 @@ Creep.prototype.transferEnergyMy = function() {
   if (!this.memory.target) {
     let structure = this.pos.findClosestByRange(FIND_MY_STRUCTURES, {
       filter: (s) => {
-        if (s.energy + s.getFutureEnergy() >= s.energyCapacity) {
+        if (s.energy >= s.energyCapacity) {
           return false;
         }
         if (s.structureType == STRUCTURE_EXTENSION) {
@@ -725,7 +725,6 @@ Creep.prototype.transferEnergyMy = function() {
       }
     } else {
       this.memory.target = structure.id;
-      structure.changeFutureEnergy(this.carry.energy);
     }
   }
 
@@ -745,8 +744,6 @@ Creep.prototype.transferEnergyMy = function() {
       // TODO Enable and check again
       this.log('transferEnergyMy: ' + returnCode + ' ' +
           target.structureType + ' ' + target.pos);
-    } else {
-      target.changeFutureEnergy(-this.carry.energy);
     }
     delete this.memory.target;
   } else {
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -595,17 +595,11 @@ Creep.prototype.construct = function() {
   if (range <= 3) {
     let returnCode = this.build(target);
     if (returnCode == OK) {
-      if (range == 1) {
-        this.move(((this.pos.getDirectionTo(target) + 3) % 8) + 1);
-      } else {
-        this.moveRandom(true);
-      }
+      this.moveRandomTowards(target.pos, 3);
       return true;
-    }
-    if (returnCode == ERR_NOT_ENOUGH_RESOURCES) {
+    } else if (returnCode == ERR_NOT_ENOUGH_RESOURCES) {
       return true;
-    }
-    if (returnCode == ERR_INVALID_TARGET) {
+    } else if (returnCode == ERR_INVALID_TARGET) {
       this.log('config_creep_resource construct: ' + returnCode + ' ' + JSON.stringify(target.pos));
       this.moveRandom();
       let filterSpawns = function(object) {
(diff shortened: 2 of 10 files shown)


src/prototype_creep_resources.js
```diff
@@ -486,7 +486,7 @@ Creep.prototype.transferMy = function() {
 };
 
 Creep.prototype.getEnergy = function() {
-  if (this.carry.energy == this.carryCapacity) {
+  if (this.carry.energy > 0) {
```
TooAngel: I don't think this will work. `getEnergy` is actually used to `harvest` energy from sources. In that case the harvester would only get 1 tick energy from the source.

src/prototype_room_my.js
```diff
@@ -262,7 +262,7 @@ Room.prototype.executeRoom = function() {
       amount = 2;
       // TODO maybe better spawn harvester when a carry recognize that the dropped energy > threshold
       if (room.controller.level == 2 || room.controller.level == 3) {
-        amount = 5;
+        amount = 3;
```
TooAngel: I think this should be removed and the carry should call harvesters on RCL < 4, when dropping energy and dropped energy > threshold.

But fine for now

src/prototype_structure.js
```diff
@@ -0,0 +1,36 @@
...
+});
+
+Structure.prototype.getFutureEnergy = function() {
+  if (this.memory.futureEnergy === undefined) {
+    return 0;
+  }
+  return this.memory.futureEnergy;
+};
+
+Structure.prototype.changeFutureEnergy = function(amount) {
```
TooAngel: I think you need to add some kind of decay if for some reason the creep will no deliver, e.g. it dies.

I started to implement a creep.died method (https://github.com/TooAngel/screeps/blob/ce4ee50d81e07f28552b3bf938653f1c2061abff/src/role_nextroomerattack.js#L24) It is not properly tested

But I think I would prefer some kind of decay to handle it on a more general level.

Keep a list of added futures with Game.time when it was added and check / remove on `getFutureEnergy` or something like that.

WhiteHalmos: +1, my idea is kind of half baked. It needs to be more robust and separate from the creep which can die or decide to do something else.

src/prototype_creep_resources.js
WhiteHalmos: Agreed, I think it's better for it to be a state system instead, where the creep gets energy until capacity, then spends energy until 0, and repeat. WDYT?

TooAngel: I tried something out with a state machine in `role_mineral`, ended in a mess :-)

Yes, sure, a state machine could make it more explicit.

src/config.js
```diff
@@ -32,7 +32,8 @@ global.config = {
   },
 
   autoattack: {
-    disabled: false
+    notify: false,
```
TooAngel: :-( Liked to see the message.

But yes, make sense, I like

WhiteHalmos:
Updated, PTAL. :)

The futures idea was a half baked implementation, didn't want to include that in the codebase.

src/prototype_creep_resources.js
```diff
@@ -1,9 +1,20 @@
...
+  let methods = [];
+
+  if (this.memory.hasEnergy === undefined) {
+    this.memory.hasEnergy = (this.carry.energy == this.carryCapacity);
+  } else if (this.memory.hasEnergy && this.carry.energy === 0) {
+    this.memory.hasEnergy = false;
+  } else if (!this.memory.hasEnergy &&
+      this.carry.energy == this.carryCapacity) {
+    this.memory.hasEnergy = true;
+  }
```
TooAngel: I would put this logic to 'getEnergy'. The task are constructed the way, that each task decides if it gets executed and nothing else is done.

So if you move this logic to the beginning of `getEnergy` and `return false`. It is the same as if you wouldn't add it.

src/prototype_creep_resources.js
```diff
@@ -12,17 +23,10 @@ Creep.prototype.harvesterBeforeStorage = function() {
...
-        return false;
-      }
-      if (object.structureType == STRUCTURE_CONTROLLER) {
-        return false;
-      }
-      return true;
+    filter: (s) => {
+      return (s.structureType != STRUCTURE_RAMPART &&
+        s.structureType != STRUCTURE_WALL &&
+        s.structureType != STRUCTURE_CONTROLLER);
```
TooAngel: I like simplifying it.

@Nevrdid started to introduce `tables` for that. I started an issue #36 maybe we could discuss and agree on a common solution.

src/prototype_creep_resources.js
TooAngel: I still think that this will prevent the harvester from getting energy out of the sources

src/prototype_creep_startup_tasks.js
```diff
@@ -463,9 +463,6 @@ Creep.prototype.repairStructure = function() {
 };
 
 Creep.prototype.getDroppedEnergy = function() {
-  if (this.carry.energy > 0.9 * this.carryCapacity) {
-    return false;
-  }
```
TooAngel: I would still keep this. Depending a bit on the getEnergy discussion.

src/prototype_creep_resources.js
WhiteHalmos: I think it should still be able to harvest. I can set it up and try.

src/prototype_creep_resources.js
WhiteHalmos: +1 in the meantime I'll leave this as is.

src/prototype_creep_startup_tasks.js
WhiteHalmos: I don't think this is necessary given the state machine, but I can keep it if you want.

src/prototype_creep_resources.js
TooAngel: Source harvesting is done in the `getEnergy` method. If the creep hasEnergy the getEnergy method will not be executed. So the creep will harvest one time, has energy, next tick, getEnergy is not executed, the creep will move away.

src/prototype_creep_startup_tasks.js
TooAngel: k

src/prototype_creep_resources.js
WhiteHalmos: Creep only hasEnergy after it is completely full. :)

PR #48, opened 2016-12-28
[WIP] Autoattack level 2

TooAngel:
squadheal and squadsiege move to the to be attacked room, will enter the room and flee if they get attacked. This drains the energy from the tower or room.
Within the room and not under attack the squadheal heals damaged creeps, the squadsiege frees up the way to the towers and spawn and kills them.

### DIFF
--- a/src/role_squadsiege.js
+++ b/src/role_squadsiege.js
@@ -40,20 +40,22 @@ roles.squadsiege.preMove = function(creep, directions) {
 
   }
 
-  if (!creep.memory.initialized) {
-    Memory.squads[creep.memory.squad].siege[creep.id] = {};
-    creep.memory.initialized = true;
-  }
-  var squad = Memory.squads[creep.memory.squad];
-  if (squad.action == 'move') {
-    if (creep.room.name == squad.moveTarget) {
-      let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);
-      let nextExit = nextExits[Math.floor(nextExits.length / 2)];
-      let range = creep.pos.getRangeTo(nextExit.x, nextExit.y);
-      if (range < 2) {
-        Memory.squads[creep.memory.squad].siege[creep.id].waiting = true;
-        creep.moveRandom();
-        return true;
+  if (creep.memory.squad) {
+    if (!creep.memory.initialized) {
+      Memory.squads[creep.memory.squad].siege[creep.id] = {};
+      creep.memory.initialized = true;
+    }
+    var squad = Memory.squads[creep.memory.squad];
+    if (squad.action == 'move') {
+      if (creep.room.name == squad.moveTarget) {
+        let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);
+        let nextExit = nextExits[Math.floor(nextExits.length / 2)];
+        let range = creep.pos.getRangeTo(nextExit.x, nextExit.y);
+        if (range < 2) {
+          Memory.squads[creep.memory.squad].siege[creep.id].waiting = true;
+          creep.moveRandom();
+          return true;
+        }
       }
     }
   }
@@ -62,6 +64,13 @@ roles.squadsiege.preMove = function(creep, directions) {
 
 //TODO need to check if it works
 roles.squadsiege.action = function(creep) {
+  if (creep.room.name !== creep.memory.routing.targetRoom) {
+    if (creep.hits < creep.hitsMax) {
+      creep.moveRanom();
+    } else {
+      delete creep.memory.routing.reached;
+    }
+  }
   creep.siege();
 };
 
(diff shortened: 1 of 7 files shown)


Somotaw:
I would suggest 

let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);

at

if (creep.memory.squad) {
        var squad = Memory.squads[creep.memory.squad];
        if (!creep.memory.initialized) {
            squad.heal[creep.id] = {};
            creep.memory.initialized = true;
        }
        let reverse = false;
        if (squad.action == 'move') {
            if (creep.room.name == squad.moveTarget) {
                **let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);**
                let nextExit = nextExits[Math.floor(nextExits.length / 2)];
                let range = creep.pos.getRangeTo(nextExit.x, nextExit.y);
                if (range < 4) {
                    Memory.squads[creep.memory.squad].heal[creep.id].waiting = true;

this fixes the undefined error after the healers reach squadtarget

src/role_squadsiege.js
```diff
@@ -40,20 +40,22 @@ roles.squadsiege.preMove = function(creep, directions) {
...
+  if (creep.memory.squad) {
+    if (!creep.memory.initialized) {
+      Memory.squads[creep.memory.squad].siege[creep.id] = {};
+      creep.memory.initialized = true;
+    }
+    var squad = Memory.squads[creep.memory.squad];
+    if (squad.action == 'move') {
+      if (creep.room.name == squad.moveTarget) {
+        let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);
+        let nextExit = nextExits[Math.floor(nextExits.length / 2)];
```
mschultheiss83: > let nextExit = nextExits[Math.floor(nextExits.length / 2)];

this is not always a good choise

src/role_squadsiege.js
```diff
@@ -40,20 +40,22 @@ roles.squadsiege.preMove = function(creep, directions) {
...
+        let range = creep.pos.getRangeTo(nextExit.x, nextExit.y);
+        if (range < 2) {
+          Memory.squads[creep.memory.squad].siege[creep.id].waiting = true;
+          creep.moveRandom();
```
mschultheiss83: > creep.moveRandom();
replace with moveRandomInRange() ;-)

src/role_squadsiege.js
```diff
@@ -62,6 +64,13 @@ roles.squadsiege.preMove = function(creep, directions) {
 
 //TODO need to check if it works
 roles.squadsiege.action = function(creep) {
+  if (creep.room.name !== creep.memory.routing.targetRoom) {
+    if (creep.hits < creep.hitsMax) {
+      creep.moveRanom();
```
mschultheiss83: same as https://github.com/TooAngel/screeps/pull/48/files#diff-4e1a69f76e2f76f6f5c95a219f74e375R56

TooAngel:
> I would suggest editing role_squadheal
> 
> let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);
> 
> at
> 
> if (creep.memory.squad) {
> var squad = Memory.squads[creep.memory.squad];
> if (!creep.memory.initialized) {
> squad.heal[creep.id] = {};
> creep.memory.initialized = true;
> }
> let reverse = false;
> if (squad.action == 'move') {
> if (creep.room.name == squad.moveTarget) {
> let nextExits = creep.room.find(creep.memory.routing.route[creep.memory.routing.routePos].exit);
> let nextExit = nextExits[Math.floor(nextExits.length / 2)];
> let range = creep.pos.getRangeTo(nextExit.x, nextExit.y);
> if (range < 4) {
> Memory.squads[creep.memory.squad].heal[creep.id].waiting = true;
> 
> this fixes the undefined error after the healers reach squadtarget

thx, fixed

TooAngel:
> please create shorthand for
> creep.memory.routing.route[creep.memory.routing.routePos], something like creep.pos eg creep.routePos() or creep.memoryRoutePos()

We could do this, I only see 5 occurrences of `routing.routePos`. I wouldn't do that in this PR, would like to concentrate on the autoattack, if I find the time.

src/role_squadsiege.js
TooAngel: The initial plan was, that the squadsiege waits ~2 fields before the exit. The path is calculated to exactly that exit, so in this case it should make sense. Any example where this is bad (in this case)?

I want to remove this anyway.

src/role_squadsiege.js
TooAngel: Hm, not sure for this case. If maybe even a `creep.moveRandom(true)` to stay on the path.

In the first version the squadsiege was flipping two to three positions before exiting the room until the squadsiege arrives, which looked promising.

Will have to redo this anyway.

src/role_squadsiege.js
TooAngel: same as the comment before :-)

PR #50, opened 2016-12-28
Skip tick if not enough cpu

Evanito:
Will skip a tick instead of using all the cpu then erroring.
Eventually it will save up enough cpu to run a tick, then become stable in only running once every several ticks, instead of erroring every single tick.


TooAngel:
Cool, thx

PR #55, opened 2016-12-29
Add external supply to queue

Tsungen:
Spawn 'reserver' after 'carry' and 'sourcer' to constantly have the
external supply lines running.
Change carryPercentage for less jumping around at one spot.

Testing on live

### DIFF
--- a/src/prototype_room_creepbuilder.js
+++ b/src/prototype_room_creepbuilder.js
@@ -94,7 +94,18 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
         }
         return 4;
       }
-
+      // add external supply in front of the other
+            if (target !== room.name) {
+              if (object.role == 'carry' ) {
+                return 4;
+              }
+              if (object.role == 'sourcer' ) {
+                return 5;
+              }
+              if (object.role == 'reserver' ) {
+                return 6;
+              }
+            }
       if (object.role == 'nextroomer') {
         return 11;
       }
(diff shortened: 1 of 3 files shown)


src/prototype_room_creepbuilder.js
```diff
@@ -94,7 +94,18 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
         }
         return 4;
       }
-
+      // add external supply in front of the other
+            if (target !== room.name) {
+              if (object.role == 'carry' ) {
+                return 4;
```
TooAngel: I think I would start at 5. Having in room creeps more important than external.

E.g. if a room is under attack, it is more important that these creeps are spawned, than an external carry. Maybe discussable

src/role_carry.js
```diff
@@ -68,13 +68,13 @@ roles.carry.preMove = function(creep, directions) {
   if (!creep.memory.routing.reverse) {
     reverse = creep.checkForTransfer(directions.forwardDirection);
   }
-
+  // less twiddling around
   let carryPercentage = 0.1;
   if (creep.room.name == creep.memory.routing.targetRoom) {
-    carryPercentage = 0.8;
+    carryPercentage = 0.5;
```
TooAngel: Maybe extract these values to `config.js`. Easier to tweak around and to overwrite.

Tsungen:
Fixed the errors and moved the queue 1 position back.
Are the comments fine?

TooAngel:
Thanks for fixing the comments, I like.

Happy to see adding the comment. I think it is good to improve the understandability (<-- is this a word?)

PR #59, opened 2016-12-30
AutoComplete folder + atom-ternjs

Nevrdid:
You must install atom-ternjs anyway for activate autocompletion. If you use ,another IDE, don't need atom-ternjs.


WhiteHalmos:
Though this is nice, perhaps we should leave editing preferences to individual devs.

Nevrdid:
yeah dunno, just did a modif on so thought about commit it. Nobody is forced to use it but it's pretty light so why not add it. Maybe better in another branch configured for atom with why not some packages settings included. But it's the only one auto-completion for screeps i think so we can't really choose another one if we want to have it.
Choice's up to you :) i already have it, just must add it in gitignore if you don't want it.

TooAngel:
I like the AutoComplete will also have a look to use it, cause I'm using Atom.

But ... :-) I would suggest to put it in a separate project:
 - It is not directly bot related
 - Maybe someone wants to use it for their own bot
 - It could be registered at npm
 - Maybe even included as Atom package ...

Nevrdid:
ok, PR closed so.

PR #60, opened 2016-12-30
modif for wait to have enouth energy before set room as active

Nevrdid:


src/prototype_room_my.js
```diff
@@ -451,7 +451,8 @@ Room.prototype.reviveRoom = function() {
   let nextRoomers = _.filter(Game.creeps, c => c.memory.role === 'nextroomer' &&
   c.memory.routing.targetRoom === this.name).length;
   if (this.controller.level >= 4 &&
-      this.controller.ticksToDowngrade > CONTROLLER_DOWNGRADE[this.controller.level] * 0.9) {
+      this.controller.ticksToDowngrade > CONTROLLER_DOWNGRADE[this.controller.level] * 0.9 &&
+      this.energyCapacityAvailable > 1000) {
```
TooAngel: Can you extract it as config value? Bonus: the downgrade factor (0.9), too

PR #68, opened 2017-01-01
Feature/attack42

mschultheiss83:


Somotaw:
I think you forgot to change brain.startMeleeSquad(closestSpawn.room, roomName to
brain.startMeleeSquad(closestSpawn.room.name, roomName
in prototype_room_attack

mschultheiss83:
please take a second look: [pseudo code]
```js
closestSpawn = this.closestSpawn(roomName);
```
and 
```js
Room.prototype.closestSpawn = function(target) {
// [...]
  return _.first(shortest).room;

```

**sorry**, need to add docs for that, _or would you_?

src/prototype_room_utils.js
```diff
@@ -1,5 +1,44 @@
...
+      pathLength[room] = {
+        room: room,
+        route: route,
+        length: routeLength
+      };
+    }
+  }
+
+  let shortest = _.sortBy(pathLength, global.utils.returnLength);
+  return _.first(shortest).room;
```
TooAngel: I think you can just use `_.min(pathLength, global.util.returnLength)`

src/prototype_room_utils.js
```diff
@@ -1,5 +1,44 @@
...
+
+  return _.sortBy(Memory.myRooms, sortByLinearDistance);
+};
+
+Room.prototype.closestSpawn = function(target) {
+  let pathLength = {};
+  let roomsMy = this.sortMyRoomsByLinearDistance(target);
+
+  for (let room of roomsMy) {
+    let route = Game.map.findRoute(room, target);
```
TooAngel: I think we should avoid occupied rooms, like this: https://github.com/TooAngel/screeps/blob/master/src/prototype_creep_routing.js#L40

src/prototype_room_utils.js
```diff
@@ -1,5 +1,44 @@
...
+    if (route && routeLength) {
+      //TODO @TooAngel please review: save found route from target to myRoom Spawn by shortest route!
+      //Memory.rooms[room].routing = Memory.rooms[room].routing || {};
+      //Memory.rooms[room].routing[room + '-' + target] = Memory.rooms[room].routing[room + '-' + target] || {
+      //    path: room + '-' + route,
+      //    created: Game.time,
+      //    fixed: false,
+      //    name: room + '-' + target,
+      //    category: 'moveToByClosestSpawn'
+      //  };
```
TooAngel: Hm, not yet sure.

So it is mainly a CPU optimization. If, it should be before `findRoute` to save CPU :-)
I don't think it will we called too often.
To properly save CPU we should store the complete route, so that the creeps could get the route from memory.
I would also suggest to change it from `Memory.rooms[room].routing` to something else to not collide with the existing, which is for in-room paths. (Better change `.routing` to `.pathing` or similar, but this would be a really big change.)

PR #69, opened 2017-01-01
Hotfix/extractor

mschultheiss83:
this is simple fix Energy in Terminal > 50%

### DIFF
--- a/src/role_extractor.js
+++ b/src/role_extractor.js
@@ -10,21 +10,69 @@ roles.extractor = {};
 
 roles.extractor.boostActions = ['harvest', 'capacity'];
 
-roles.extractor.getPartConfig = function(room, energy, heal) {
+roles.extractor.getPartConfig = function (room, energy, heal) {
   var parts = [MOVE, CARRY, MOVE, WORK];
   return room.getPartConfig(energy, parts);
 };
 
-roles.extractor.energyBuild = function(room, energy, heal) {
+roles.extractor.energyBuild = function (room, energy, heal) {
   var max = 2000;
   energy = Math.max(250, Math.min(max, room.getEnergyCapacityAvailable()));
   return energy;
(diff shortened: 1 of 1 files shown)


src/role_extractor.js
```diff
@@ -21,10 +21,58 @@ roles.extractor.energyBuild = function(room, energy, heal) {
   return energy;
 };
 
-roles.extractor.action = function(creep) {
-  return creep.handleExractor();
-};
+roles.extractor.terminalStorageExchange = function(creep) {
+  var terminal = creep.room.terminal;
+  if (!terminal || !terminal.isActive()) {
+    // TODO kill creep?
```
TooAngel: Killing the creep is fine.

The `isActive()` method is somehow expensive, could be fine for just the mineral roles

src/role_extractor.js
```diff
@@ -21,10 +21,58 @@ roles.extractor.energyBuild = function(room, energy, heal) {
...
+    if (creep.withdraw(terminal, RESOURCE_ENERGY) !== OK) {
+      creep.moveTo(terminal);
+    }
+  }
 
-roles.extractor.execute = function(creep) {
-  return creep.handleExractor();
+  if (!action.withdraw || action.transfer) {
+    if (creep.transfer(creep.room.storage, RESOURCE_ENERGY) !== OK) {
+      creep.moveTo(creep.room.storage);
```
TooAngel: I guess this will run into problems. Especially because the harvester creep is on the path and block the moveTo, because no `ignoreCreeps`. If `ignoreCreeps` is set, it could get stuck at the filler or builder creep.

But it also can work sometimes ...

TooAngel:
Thanks for the PR, I think the comments are not so important, so I would merge it and the issues can be addresses later.

PR #75, opened 2017-01-02
fix nextroomer

Tsungen:
they should fill energy now before upgrading the controller

### DIFF
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -205,7 +205,7 @@ Room.prototype.handleScout = function() {
   }
   let shouldSpawn = (
     ((Game.time + this.controller.pos.x + this.controller.pos.y) %
-        config.room.scoutInterval) === 0 &&
+      config.room.scoutInterval) === 0 &&
     this.controller.level >= 2 &&
     this.memory.queue.length === 0 &&
     config.room.scout
@@ -462,9 +462,9 @@ Room.prototype.reviveRoom = function() {
   let nextRoomers = _.filter(Game.creeps, c => c.memory.role === 'nextroomer' &&
     c.memory.routing.targetRoom === this.name).length;
   if (this.controller.level >= 4 &&
-      this.controller.ticksToDowngrade >
-       (CONTROLLER_DOWNGRADE[this.controller.level] * config.nextRoom.minDowngradPercent / 100) &&
-      this.energyCapacityAvailable > config.nextRoom.minEnergyForActive) {
+    this.controller.ticksToDowngrade >
+    (CONTROLLER_DOWNGRADE[this.controller.level] * config.nextRoom.minDowngradPercent / 100) &&
+    this.energyCapacityAvailable > config.nextRoom.minEnergyForActive) {
 
     this.memory.active = true;
     return false;
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -242,7 +242,10 @@ Room.prototype.executeRoom = function() {
 
   if (spawns.length === 0) {
     this.reviveRoom();
-  } else if (this.energyCapacityAvailable < 1000) {
+  }
+  // Revive Room if energycapacity or storage don't met minimum
+  else if (this.energyCapacityAvailable < 1000 ||
+  this.storage.store.energy < config.nextRoom.minStorage) {
     this.reviveRoom();
     if (hostiles.length > 0) {
       this.controller.activateSafeMode();
@@ -462,9 +465,10 @@ Room.prototype.reviveRoom = function() {
   let nextRoomers = _.filter(Game.creeps, c => c.memory.role === 'nextroomer' &&
     c.memory.routing.targetRoom === this.name).length;
   if (this.controller.level >= 4 &&
-    this.controller.ticksToDowngrade >
-    (CONTROLLER_DOWNGRADE[this.controller.level] * config.nextRoom.minDowngradPercent / 100) &&
-    this.energyCapacityAvailable > config.nextRoom.minEnergyForActive) {
+      this.controller.ticksToDowngrade >
+       (CONTROLLER_DOWNGRADE[this.controller.level] * config.nextRoom.minDowngradPercent / 100) &&
+      this.energyCapacityAvailable > config.nextRoom.minEnergyForActive &&
+      this.storage.store.energy > config.nextRoom.minStorage) {
 
     this.memory.active = true;
     return false;
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -244,7 +244,7 @@ Room.prototype.executeRoom = function() {
     this.reviveRoom();
   }
   // Revive Room if energycapacity or storage don't met minimum
-  else if (this.energyCapacityAvailable < 1000 ||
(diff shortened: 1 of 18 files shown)


TooAngel:
Hm, I'm not so sure about that.

I think we should usually upgrade the room to RCL 4 as fast as we can.
So build up the room to RCL3 to have a tower for defense, start building the spawn and tower (or other buildings), but continue to upgrade the controller. With RCL 4 the storage is available and the better energy distribution can drop in.
Maybe we could make some change for lower GCL. On higher (maybe 3 or 4) the nextroomers are big creeps and the room itself can't spawn anything comparable. And if it starts to spawn, the logic needs to get much more complicated.
I had a lot of cases where two single WORK part creeps where constantly blocking the sources getting nearly no energy out of the source, while the 50 part nextroomer standing behind them.

If you still think it can make sense, I guess there needs to be improvement on the spawning logic of a room in reviving mode and I would only do `transferEnergy` at RCL 4, maybe, maybe ... maybe RCL 3, with goods arguments.

WhiteHalmos:
Definitely prefer using nextroomers instead of single or double work part "sourcers" blocking sources (which I've seen).

Tsungen:
I will test it again with my controller at rcl4.
With the actual logic i think it will still upgrade until rcl8 without transfering any energy

TooAngel:
Rooms start to spawn creeps properly if no nextroomer is in the room or the RCL > 4. (https://github.com/TooAngel/screeps/blob/master/src/prototype_room_my.js#L264)

Nextroomers are only send if there is no spawn or the `room.energyCapacityAvailable < 1000` (https://github.com/TooAngel/screeps/blob/master/src/prototype_room_my.js#L243)

The `RCL < 8` check is only for the case that a room was destroyed and we are reviving it. Nextroomer shouldn't upgrade the controller beyond RCL 8.

If you have an issue with a room, you can quickly paste it in slack so that I can have a look.

src/prototype_room_my.js
```diff
@@ -242,7 +242,10 @@ Room.prototype.executeRoom = function() {
 
   if (spawns.length === 0) {
     this.reviveRoom();
-  } else if (this.energyCapacityAvailable < 1000) {
+  }
+  // Revive Room if energycapacity or storage don't met minimum
+  else if (this.energyCapacityAvailable < 1000 ||
+  this.storage.store.energy < config.nextRoom.minStorage) {
```
TooAngel: I sometimes ran into the case that the room storage was empty - Usually messed up something, recovers after some time. This would trigger nextroomers where it is not necessary / helpful.

But maybe I shouldn't mess up stuff :-) And if it happens I should fix it and get reminded by seeing the nextroomers :-) (Or I change the config to 0).

So I'm fine with it, just want to mention it.

src/prototype_room_my.js
```diff
@@ -242,7 +242,10 @@ Room.prototype.executeRoom = function() {
 
   if (spawns.length === 0) {
     this.reviveRoom();
-  } else if (this.energyCapacityAvailable < 1000) {
+  }
+  // Revive Room if energycapacity or storage don't met minimum
+  else if (this.energyCapacityAvailable < 1000 ||
```
TooAngel: `this.energyCapacityAvailable < minEnergyForActive` ?!

src/role_nextroomer.js
```diff
@@ -308,11 +308,15 @@ roles.nextroomer.settle = function(creep) {
     methods.push(Creep.constructTask);
   }
 
+  // Fill energy after building is done
+  if (creep.room.controller.level >= 4) {
+    methods.push(Creep.transferEnergy);
+  }
```
TooAngel: As mentioned in another comment. We should also consider the case:
RCL (5-8), destroyed structures, room needs to be revived. Are the nextroomers busy with filling the structures?
.... _thinking_ ... Should be fine.

Just want to make you aware of other situations.

src/prototype_room_my.js
Tsungen: In my case i had no storage and a 300 Energy Harvester running around getting nothing done.
Room was set to active.
With the `energyFromStorageThreshold: 2000` it shouldn't trigger revive until needed.

src/role_nextroomer.js
Tsungen: As long as there is something to build they will build. For RCL8 this will take some time.

src/prototype_room_my.js
Nevrdid: `this.energyCapacityAvailable < config.nextRoom.minEnergyForActive`

aceofaces:
I would say that RCL 3 would be considered a good point to start storing but a check if any of my rooms are over RCL 4 would be even better.  Once the newroomer spawning room becomes RCL >= 5 you should be able to use more of the rooms energy than it produces upgrading the controller instead of on producing creeps.

In a restart situation with the main room at RCL 4, the nextroomers harvest as fast as a standard sourcer, making the chain of sourcer, carry, drop for next roomers to rush upgrade and get full utilization out of both sources better than having nextroomers tap one source and line up for their time.

Flat setting it to RCL 3 would be a beginning or edge case recovery improvement, but a detriment for later RCL level newroomers.

src/role_nextroomer.js
aceofaces: I think I found a solution that would base filling upon how many parts the creep has.
If the nextroomer would be only as efficient as a sourcer it has been better in my testing to be building sourcers and having carries deliver to the nextroomers to keep more time on upgrading the controller.

 ```
 // Fill energy after building is done
  if (creep.getActiveBodyparts('work') <= 5 && creep.room.controller.level >= 3) {
    methods.push(Creep.transferEnergy);
  }
```

Nevrdid: working on a script for transfer energy from hight stored energy room to lowest one. i think it can resolve all energy problems at start.

TooAngel: > I think I found a solution that would base filling upon how many parts the creep has.
> If the nextroomer would be only as efficient as a sourcer it has been better in my testing to be building sourcers and having carries deliver to the nextroomers to keep more time on upgrading the controller.
> 
> // Fill energy after building is done
>  if (creep.getActiveBodyparts('work') <= 5 && creep.room.controller.level >= 3) {
>    methods.push(Creep.transferEnergy);
>  }

I like the approach. If you add that you should also make sure that the rooms starts to spawn sourcer.

Maybe you could even make a direct check `nextroomer.work.parts < sourcer.would.have.work.parts`.
This could be done by executing the `role.sourcer.getPartConfig` method and filter the work parts.

https://github.com/TooAngel/screeps/blob/master/src/role_sourcer.js#L22

TooAngel:
> working on a script for transfer energy from hight stored energy rooms to lowest ones. i think it can resolve all energy problems at start.

I like that, too.

TooAngel:
@Tsungen Could you please merge master or rebase onto master. Currently it is tricky to find the actual PR related changes in the diff.

Tsungen:
I did rebase but that branch seems to be lost.
Did some mistakes early and cant see where.
Will open a new PR.

TooAngel:
@Tsungen Can we close this PR?

PR #77, opened 2017-01-02
[WIP] Role: Extractor: Updated docs

mschultheiss83:
plaese leave comments how to integrade an update on creep.memory._move path by using  
- a custom layout for the room 
- and try not using current pathes

### DIFF
--- a/src/role_extractor.js
+++ b/src/role_extractor.js
@@ -23,8 +23,16 @@ roles.extractor.energyBuild = function(room, energy, heal) {
 
 roles.extractor.terminalStorageExchange = function(creep) {
   var terminal = creep.room.terminal;
-  if (!terminal || !terminal.isActive()) {
-    // TODO kill creep?
+  /**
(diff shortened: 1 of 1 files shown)


mschultheiss83:
i ve got an bug on role_extractor.js:85 ff

i'm just on ~ 0.5 capacity of the storage, creep puts in and takes out, puts in and takes out, and so on... some ideas?

TooAngel:
Regarding the energy put in take out:
https://github.com/TooAngel/screeps/blob/master/src/role_harvester.js#L40

I guess something similar to this will work. Calculated based on `creep.carry.energy + storage.store`.
Didn't look deeper in it, but that was usually the case.

TooAngel:
Started an issue, so that we can discuss and form our ideas: https://github.com/TooAngel/screeps/issues/79

mschultheiss83:
closed, see #79
and https://github.com/TooAngel/screeps/pull/111

PR #78, opened 2017-01-03
README and checkRoleToSpawn for sourcer and reserver 

TooAngel:
Sourcer and server can queue up way too much, using checkRoleToSpawn prevents this.

Split out the AI desgin to a separate file.

### DIFF
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -839,15 +839,7 @@ Creep.prototype.handleReserver = function() {
 
       if (resource_structures.length > 0 && !creep.room.controller.my) {
         creep.log('Call structurer from ' + creep.memory.base + ' because of ' + resource_structures[0].structureType);
-        Game.rooms[creep.memory.base].memory.queue.push({
-          role: 'structurer',
-          routing: {
-            targetRoom: this.name,
-            reached: false,
-            routePos: 0,
-            pathPos: 0
-          }
-        });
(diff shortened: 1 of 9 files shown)


src/prototype_creep_resources.js
```diff
@@ -874,21 +874,12 @@ Creep.prototype.handleReserver = function() {
       });
 
       if (sourcer.length < sources.length) {
-        for (var sources_id in sources) {
-          if (checkSourcerMatch(sourcer, sources[sources_id].pos)) {
+        for (let source of sources) {
```
TooAngel: Thx, changed.

src/utils.js
```diff
@@ -138,6 +138,12 @@ global.utils = {
 
     console.log(JSON.stringify(minerals));
     console.log(minerals.U);
+  },
+
+  queueCheck: function(roomName) {
+    console.log(JSON.stringify(_.countBy(Game.rooms[roomName].memory.queue), function(object) {
+      return object.role;
+    }));
```
mschultheiss83: create: 
```js
let prop = function(prop) { 
  return function (object) { 
    return object[prop]; 
  }
}
```
for improvement:
```js
var found = _.countBy(Memory.rooms[roomName].queue), prop('role') 
console.log(JSON.stringify(found));
return found;
```

TooAngel: Not sure, added it like you suggested.

Issue #80, opened 2017-01-03
feature/Add screeps-visual and make visualizer manager object


Somotaw:
- for visualizing different paths or other stuff
when I did this it also showed routes that look way off, perhaps routing needs some work? :)

Somotaw:
got this done, but I cannot create pull requests so i'll just link it here.
https://github.com/Somotaw/screeps

WhiteHalmos:
Pictures please? I don't have screeps-visual. :)

Somotaw:
![image](https://cloud.githubusercontent.com/assets/6491698/21599438/cda4e4c8-d16c-11e6-88f8-2745f918e6a3.png)

Somotaw:
also try the fork and you do ;)

Somotaw:
whoever made visual.js needs to work on their === and ; XD

Somotaw:
maybe this is cause for cpu hog?

Somotaw:
![image](https://cloud.githubusercontent.com/assets/6491698/21600407/f80dd2c4-d17b-11e6-8b2f-e863aea7cf02.png)
res is path's for creeps and while is for rooms in my setup

TooAngel:
I think it is an issue with the screeps-visual. Most of the creeps only move on the pre calculated paths, if they were off the creeps were stuck at the walls and not try to find another way.

As always, I could be wrong ;-)

TooAngel:
> maybe this pathing is the cause of the crazy cpu im getting testing this ^^

Not sure if you already recognize the screeps-profiler integration, I think this is good for debugging the CPU issues.

Somotaw:
Will test the current function in a controlled enviroment and come back with definite answers.

Yes I've been using screeps-profiler alot, great tool :)

Somotaw:
Fixed som stuff on local branch, the deviant lines in my example are based off direct room memory, once I used the getMemoryPaths all the roompaths now look ok :)

Somotaw:
![image](https://cloud.githubusercontent.com/assets/6491698/21645444/465afed6-d292-11e6-851c-8ab3da0431b8.png)

Somotaw:
starts to look pretty good eand easy to visualize building layout and pathfinding for squads with this :D

TooAngel:
The implementation looks promising.
What is currently missing from my perspective is some kind of documentation.
Like:
 - Which browser script do I need to install
 - How to enable
 - What are the config flags


[...]

PR #81, opened 2017-01-03
Add screeps-visual and visualize creep paths from _move and room paths from room cache.

Somotaw:


Somotaw:
hmm CircleCl does not like Visual.js from that project heheh

Somotaw:
Should be ready to merge. run with the global loadVisual() after loading screeps in browser or steam client.

src/main.js
```diff
@@ -1,42 +1,44 @@
...
+    return;
+  }
+
+  brain.prepareMemory();
+  brain.handleNextroom();
+  brain.handleSquadmanager();
+  brain.handleIncomingTransactions();
+  Memory.myRooms = _.map(_.filter(Game.rooms, (r) => r.execute()), r => r.name);
+  global.visualizePaths();
+  global.RawVisual.commit();
```
TooAngel: Would encapsulate it with a config flag, `visualsomething`, default to false for bot deployments.

src/main.js
```diff
@@ -1,42 +1,44 @@
...
+    console.log('Skipping tick ' + Game.time + ' due to lack of CPU.');
+    return;
+  }
+
+  brain.prepareMemory();
+  brain.handleNextroom();
+  brain.handleSquadmanager();
+  brain.handleIncomingTransactions();
+  Memory.myRooms = _.map(_.filter(Game.rooms, (r) => r.execute()), r => r.name);
+  global.visualizePaths();
```
TooAngel: Do we have to execute it here, or is it possible to execute it as a console command? I saw `visual.js` didn't have a proper look at it.
My idea would be to have some commands for the console: `visual.showPaths('E8S8')`, `visual.showStructures('E8S8')` ... or something like this. So not loading it always, but on demand.
I guess if I deploy this to the TooAngel account, I will run into CPU issues for 22 rooms.

Not sure if it makes sense or possible with visual.

src/main.js
```diff
@@ -1,42 +1,44 @@
...
+
+module.exports.loop = function() {
+  if (config.profiler.enabled) {
+    profiler.wrap(function() {
+      main();
+    });
+  } else {
+    main();
+  }
+};
```
TooAngel: Not sure why the complete file changed, maybe line breaks? If so: No :-)

src/visual.js
```diff
@@ -0,0 +1,358 @@
+'use strict';
```
TooAngel: If this is the visual.js from https://github.com/screepers/screeps-visual I would suggest to add it via `package.json` and copy it in the grunt task. Like with `screeps-profiler`:
 - https://github.com/TooAngel/screeps/blob/master/package.json#L21
 - https://github.com/TooAngel/screeps/blob/master/Gruntfile.js#L132
 - https://github.com/TooAngel/screeps/blob/master/src/main.js#L10

src/visualize_paths.js
```diff
@@ -0,0 +1,79 @@
...
+//jscs:disable requireSpaceBetweenArguments
+//jscs:disable requireSpacesInConditionalExpression
+//jscs:disable requireSpaceAfterBinaryOperators
+//jscs:disable disallowSpaceAfterPrefixUnaryOperators
+//jscs:disable validateParameterSeparator
+//jscs:disable requireSpaceBeforeBlockStatements
+//jscs:disable disallowSpacesInsideObjectBrackets
+//jscs:disable disallowSpacesInAnonymousFunctionExpression
+//jscs:disable disallowMultipleLineBreaks
+//jscs:disable disallowSpacesInFunctionExpression
```
TooAngel: Nice try ;-)

If you have a good reason to disable one of the checks I'm fine, but disabling all of them should be avoided.

src/visualize_paths.js
```diff
@@ -0,0 +1,79 @@
...
+  let COLOR_PATH = colors.push('rgba(255,255,255,0.5)') - 1;
+  _.each(Game.rooms, (room, name) => {
+    let visual = new Visual(name);
+    visual.defineColors(colors);
+    visual.setLineWidth = 0.5;
+    _.each(Game.creeps, creep=> {
+      if (creep.room != room) return;
+      let mem = creep.memory;
+      if (mem._move) {
+        let path = Room.deserializePath(mem._move.path);
```
TooAngel: Not sure what this is about. Never looked into `_move` and especially not `_move.path`. I'm only using the `room.memory.routing` paths.

Could be fine, don't know.

src/visualize_paths.js
```diff
@@ -0,0 +1,79 @@
...
+        if (path.length) {
+          //console.log('drawing route ' + mem._move.path);
+          visual.drawLine(path.map(p=>([p.x, p.y])), COLOR_RED, { lineWidth: 0.1 });
+        }
+      }
+    });
+    if (room.memory.routing)
+      for (var route in room.memory.routing) {
```
TooAngel: Just as an idea, you could use: https://github.com/TooAngel/screeps/blob/master/src/prototype_room_memory.js

This would use cached paths additional to the fixed ones and manages the deserialization for you.

TooAngel:
I would also suggest to squash the commits before merging. (https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)

Also you can add 'Fixes #80' in the commit message and the issue will be closed as soon as the PR is merged.

Somotaw:
Pretty new to js and github, but i'll work on this :) Need to sort code formatting etc, esp with all those tests ;). Been up too long i'll work more on this after I sleep :)

TooAngel:
> Pretty new to js and github, but i'll work on this :) Need to sort code formatting etc, esp with all those tests ;). Been up too long i'll work more on this after I sleep :)

:-) No, issue. Thinking about adding as a task for 'The Angels' alliance: "Improving our collaborative coding skills", I think we all are learning a lot by playing this game and working together on our repository.

If you have any question, doubts, ... you can always ping me or ask in #angels on slack of gitter if you like. I'm happy to help.

src/main.js
Somotaw: This should be doable :) and yes I suspect you will :/ I will work on making this more on demand.

src/main.js
Somotaw: Already done on local copy, will push this soon.

src/main.js
Somotaw: Researching this now.

src/visualize_paths.js
Somotaw: Temporary fix ;)

src/visual.js
Somotaw: Yeah this option sounds way better, just not used to the common workflow :)

src/visualize_paths.js
Somotaw: I believe having both as an option will be good for debugging, there are clearly path's assigned ._move maybe its just me being new to screeps but I will look into the docs to see if this is the functionality I was looking for or if I should only watch room.memory.routing.

Somotaw:
Ahh hehe ^^ I'm sure this is a worthwhile experience for everybody :) 

Thanks, will do 👍

Somotaw:
hmm squash does not work for me here XD

src/main.js
Somotaw: no trailing whitespaces or anything, hmm

Somotaw:
TODO: somehow disable tests for screeps-visual? Ideas?

src/main.js
TooAngel: I guess you are working on windows and you have the '\r\n\' while the project just have '\n'. 

Something like this: http://stackoverflow.com/questions/17579553/windows-command-to-convert-unix-line-endings
but in the other direction :-)

Somotaw:
Does anyone have any ideas for additions/changes before I commit myself to new issues?

src/main.js
```diff
@@ -1,11 +1,16 @@
 'use strict';
 
+var visualizer = {
+  render: function() { }
+};
```
TooAngel: Is this really needed?

src/main.js
```diff
@@ -22,13 +27,15 @@ var main = function() {
...
   }
+
   brain.prepareMemory();
   brain.handleNextroom();
   brain.handleSquadmanager();
   brain.handleIncomingTransactions();
-
   Memory.myRooms = _.map(_.filter(Game.rooms, (r) => r.execute()), r => r.name);
-
+  if (config.visualizer.enabled && config.visualizer.refresh) {
```
TooAngel: `&& visualizer`

To be extra save and can avoid the definition before?!

src/visualizer.js
```diff
@@ -0,0 +1,114 @@
+'use strict';
+
+var RawVisual = {
+  commit: function() { }
+};
+if (config.visualizer.enabled) {
+  try {
+    var Visual = require('visual');
```
TooAngel: we usually use lower case variables => `visual`

src/visualizer.js
```diff
@@ -0,0 +1,114 @@
+'use strict';
+
+var RawVisual = {
+  commit: function() { }
+};
```
TooAngel: I would guess this isn't need, too. But I could be wrong.

src/visualizer.js
Somotaw: This is fixed on local. Renamed screepsVisual.

src/visualizer.js
Somotaw: Wont get through jshint unless I disable warnings or forward declare.

>    src/main.js
>      36 |  if (config.visualizer.enabled && config.visualizer.refresh && visualizer) {
>                                                                          ^ 'visualizer' is not defined.
>      37 |    visualizer.render();
>              ^ 'visualizer' is not defined.
>    src/visualizer.js
>      79 |      RawVisual.commit();
>                ^ 'RawVisual' is not defined.
>      91 |      RawVisual.commit();
>                ^ 'RawVisual' is not defined.
>     103 |      RawVisual.commit();
>                ^ 'RawVisual' is not defined.

src/main.js
Somotaw: does not make any difference :/

src/main.js
Somotaw: see comment further down

src/main.js
TooAngel: I think you don't have do define any mocking objects, just make sure they are not called if not available.

src/main.js
TooAngel: see comment further down :-)

.jshintrc
```diff
@@ -1,135 +1,137 @@
...
+		"config": false,
+		"brain": false,
+		"roles": false,
+		"friends": false,
+		"ORDER_SELL": false,
+		"SUBSCRIPTION_TOKEN": false,
+		"cache": false,
+		"visualizer": false,
+		"RawVisual": false
+	}
```
TooAngel: Hm, would prefer to keep the previous two space formatting

PR #90, opened 2017-01-05
fix structurer

Tsungen:
because we start external at RCL4 we should lower the energy to build em


WhiteHalmos:
+1, move to config along with the reservation 1300 number?

PR #91, opened 2017-01-05
Screepsplus adaptation

Nevrdid:
Must enable stats & uncomment token in config.stats for use include screepsplus agent.
Datas send by default to shared tooangels account for parse scripts executions.
You can change token or comment it if you prefer the external agent.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -36,7 +36,7 @@ global.config = {
     enabled: false,
     summary: false,
     //--- Uncomment screepsplusToken for share datas on TooAngels datasboard on screepspl.us
-    //--- You can acces datas contacting tooangel on tooangels slack channel or use your token.
+    //--- You can acces datas contacting tooangel on #the_angels slack channel or use your token.
     //screepsplusToken: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6InRvb2FuZ2VscyIsImlhdCI6MTQ4MzU2MTU3OSwiYXVkIjoic2NyZWVwc3BsLnVzIiwiaXNzIjoic2NyZWVwc3BsLnVzIn0.NhobT7Jg8bOAg-MYqrYsgeMgXEVXGVYG9s3G9Qpfm-o'
   },
 
--- a/src/config.js
+++ b/src/config.js
@@ -35,6 +35,9 @@ global.config = {
   stats: {
     enabled: false,
     summary: false,
+    //--- Uncomment screepsplusToken for share datas on TooAngels datasboard on screepspl.us
+    //--- You can acces datas contacting tooangel on tooangels slack channel or use your token.
+    //screepsplusToken: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6InRvb2FuZ2VscyIsImlhdCI6MTQ4MzU2MTU3OSwiYXVkIjoic2NyZWVwc3BsLnVzIiwiaXNzIjoic2NyZWVwc3BsLnVzIn0.NhobT7Jg8bOAg-MYqrYsgeMgXEVXGVYG9s3G9Qpfm-o'
   },
 
   autoattack: {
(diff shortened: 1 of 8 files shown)


src/config.js
```diff
@@ -29,6 +29,9 @@ global.config = {
   stats: {
     enabled: false,
     summary: false,
+    //--- Uncomment screepsplusToken for share datas on TooAngels datasboard on screepspl.us
+    //--- You can acces datas contacting tooangel on tooangels slack channel or use your token.
```
TooAngel: 'tooangels slack channel' => '#the_angels slack channel'

src/config_brain_memory.js
```diff
@@ -108,8 +106,24 @@ brain.prepareMemory = function() {
...
+      Memory.stats[name + '.cpu.bucket'] = Game.cpu.bucket;
+      Memory.stats[name + '.exec.halt'] = Game.cpu.bucket < Game.cpu.tickLimit * 2;
 
+      Memory.stats[name + '.gcl.level'] = Game.gcl.level;
+      Memory.stats[name + '.gcl.progress'] = Game.gcl.progress;
+      Memory.stats[name + '.gcl.progressTotal'] = Game.gcl.progressTotal;
+
+      Memory.stats[name + '.rooms.available'] = Game.rooms.length;
+    }
+  }
```
TooAngel:
```
} else {
    delete Memory.stats;
}
```

No necessary, just an idea.

ags131: Memory.stats can also have nested objects to make it easier, ex:
```javascript
Memory.stats[name] = {
  cpu:{
    limit: Game.cpu.limit,
    tickLimit: Game.cpu.tickLimit,
    bucket: Game.cpu.bucket
  },
  exec: {
    halt: Game.cpu.bucket < Game.cpu.tickLimit * 2
  },
  gcl:{
    level: Game.gcl.level,
    progress: Game.gcl.progress,
    progressTotal: Game.gcl.progressTotal
  },
  rooms: {
    available: Game.rooms.length
  }
}
```

TooAngel: Does this work for the node-agent, too?

Previous I was using some other screeps-stats thingy (don't remember the name anymore), where it wasn't possible.

ags131: Yes, originally, the screeps stats programs didn't allow it, shortly after node-agent started using it it got merged back into the other projects

Tsungen:
Will this work on private sever?

Somotaw:
I beleiave you have to modify your private server either with mods or in source before build in order to enable the password to be set for the token and the request url to be changed

Nevrdid:
"This code was tested in the public server on Firefox, Chrome and the Steam client. Steam client private server tests were successful too." .Semperrabbit

Tsungen:
error with clean memory
`[09:15:08]TypeError: Cannot read property 'controller' of undefined
    at Object.add (stats:18:55)
    at Object.addRoot (stats:46:10)
    at main (main:36:9)
    at Object.wrap (screeps-profiler:324:16)
    at Object.module.exports.loop (main:46:14)
    at __mainLoop:1:52`

src/main.js
```diff
@@ -6,7 +6,9 @@ require('prototype_creep_move');
 require('prototype_roomPosition');
 require('prototype_room_init');
 require('prototype_room_costmatrix');
-require('visualizer');
+require('screepsplus');
+
+var Stats = require('stats');
```
TooAngel: I think we should adding it to `brain` is better than storing it in a local variable. Main motivation for that is simplifying the minification.
Would like to be able to minify the code for bot release. Also the are rumors that it saves CPU on the live server to have only one file.

src/main.js
```diff
@@ -30,7 +30,13 @@ var main = function() {
   brain.handleNextroom();
   brain.handleSquadmanager();
   brain.handleIncomingTransactions();
+
+<<<<<<< HEAD
+  //Memory.myRooms = _.create(Room.prototype, Memory.myRooms);
+=======
   //Memory.myRooms = _.create(Room.prototype, Memory.myRooms);  //thinking about save all rooms datas : -CPU => +Memory
+>>>>>>> 9191a56... change visualizer render place
+
```
Somotaw: This is what happens to mee aswell when I rebase, maybe it has something to do with removing old commit messages, sometimes that seems to break the commit, not sure what causes it yet, meaybe some of you git gurus could shine some light on this? :D

Issue #84, opened 2017-01-04
Sourcer and Carry Creep Role Failed


aceofaces:
Regression occurred sometime recently.  A single room works fine, but once external rooms need sourcers and carry they freeze in position once on the main path.

Clearing creep memory does nothing to resolve, clearing the main rooms memory does resolve the issue for a period of time.

Stack traces:
http://pastebin.com/hSagsb3G
http://pastebin.com/iRHvrnmL

aceofaces:
This issue occurs when manually sending a reserver to a neighboring room when the main room has a misplaced spawn.  Once the misplaced spawn is resolved the issue no longer occurs. 
I rolled back to 1.1.12 and the issue was still happening.  My misplaced spawn is fixed now so I can't test further.

TooAngel:
Could see the error on the first looks, quickly copying the important line here:

```
16502814 W6S77  sourcer-310          [room W6S77 pos 28,23] Executing creep role failed: W6S77 sourcer-310 586c4a5b01d50d92200cc243 {"x":28,"y":23,"roomName":"W6S77"} TypeError: Cannot read property 'x' of undefined
TypeError: Cannot read property 'x' of undefined
    at Object.roles.sourcer.preMove (role_sourcer:35:24)
```

Can be tested and fixed on a private server.

TooAngel:
You can add "#Fixes 84" to the commit message and if that commit gets merged, the issue is automatically closed.

PR #92, opened 2017-01-06
Forced Reserver Sourcer & Carry Fix

aceofaces:
Fix for https://github.com/TooAngel/screeps/issues/84

Use 'target' instead of 'x/y' or 'pos'

### DIFF
--- a/src/role_carry.js
+++ b/src/role_carry.js
@@ -37,8 +37,8 @@ roles.carry.preMove = function(creep, directions) {
       var source = creep.room.memory.position.creep[targetId];
       // TODO better the position from the room memory
       if (source !== null) {
-        let returnCode = creep.moveTo(targetId);
-        if (creep.pos.getRangeTo(targetId) > 1) {
+        let returnCode = creep.moveTo(Game.getObjectById(targetId));
+        if (creep.pos.getRangeTo(Game.getObjectById(targetId)) > 1) {
           return true;
         }
       }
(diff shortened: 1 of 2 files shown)


src/role_carry.js
```diff
@@ -37,8 +37,8 @@ roles.carry.preMove = function(creep, directions) {
       var source = creep.room.memory.position.creep[targetId];
       // TODO better the position from the room memory
       if (source !== null) {
-        let returnCode = creep.moveTo(source.pos);
-        if (creep.pos.getRangeTo(source.pos) > 1) {
+        let returnCode = creep.moveTo(targetId);
+        if (creep.pos.getRangeTo(targetId) > 1) {
```
Somotaw: Use Game.getObjectById(targetID) I dont believe moveTo can take Id numbers directly, either position or object.
Bonus points if we can use saved paths here?

src/role_sourcer.js
```diff
@@ -32,8 +32,8 @@ roles.sourcer.preMove = function(creep, directions) {
 
     var source = creep.room.memory.position.creep[targetId];
     // TODO better the position from the room memory
-    creep.moveTo(source.x, source.y);
-    if (creep.pos.getRangeTo(source.x, source.y) > 0) {
+    creep.moveTo(targetId);
+    if (creep.pos.getRangeTo(targetId) > 0) {
```
Somotaw: Same as above.

aceofaces:
That worked much better, thanks @Somotaw

TooAngel:
Thanks, I like.

PR #95, opened 2017-01-06
stats and screepsplus

Nevrdid:


src/prototype_room_my.js
```diff
@@ -346,6 +346,35 @@ Room.prototype.executeRoom = function() {
     }
   }
 
+  let carryHelpInterval = config.nextRoom.ticksUntilHelpCheck;
+  if (Game.time % carryHelpInterval) {
+    this.memory.Energy += this.energyAvailable;
```
TooAngel: `this.memory.Energy` => `this.memory.energyAvailableSum` or something like that.

src/prototype_room_my.js
```diff
@@ -346,6 +346,35 @@ Room.prototype.executeRoom = function() {
...
+  } else if (!this.memory.Energy) {
+    this.memory.Energy = 0;
+  } else {
+    let storage = this.storage;
```
TooAngel: I think you can just use `this.storage` where currently `storage` is used.

src/prototype_room_my.js
```diff
@@ -346,6 +346,35 @@ Room.prototype.executeRoom = function() {
...
+    if (this.memory.Energy < 300 * carryHelpInterval) {
+      Memory.needEnergy.push(this.name);
```
TooAngel: This isn't initialized `Memory.needEnergy`

src/prototype_room_my.js
```diff
@@ -346,6 +346,35 @@ Room.prototype.executeRoom = function() {
...
+      console.log('!!!', this.name, ' need energy !!!');
+    } else if (this.memory.Energy > 1000 * carryHelpInterval && storage.length) {
+      this.memory.queue.push({
```
TooAngel: Would use `this.checkRoleToSpawn('carry', 1, storage[0].id, this.name)` instead.

PR #97, opened 2017-01-07
stats and screepsplus

Nevrdid:


Nevrdid:
https://github.com/TooAngel/screeps/pull/91

src/screepsplus.js
```diff
@@ -0,0 +1,40 @@
...
+    };
+    document.pushStats();
+    setInterval(document.pushStats, 15000);
+  }
+  </SCRIPT>`;
+  console.log(output.split('\n').map((s)=>s.trim()).join(''));
+};
+if (config.stats.enabled && config.stats.screepsplusToken) {
+  runAgent();
+}
```
Somotaw: Suggest adding the screepsplus agent as a submodule or npm if it has the correct js file, not sure. then edition gruntfile to include it in dist :)

Nevrdid: It's only one module. dunno if it's so interesting to use external way. Plus i don't know how to do :)

PR #98, opened 2017-01-07
Manage lowenergy rooms

Nevrdid:
A must have :)
Calculate energyAvailable along congifuredpediode then set them as helper or neddHelp. Then, each 100 ticks end, if a room have a too hight energy average, she will spawn a carry for send energy to nearest one which need it.
If a needHelp room have hostile creeps, a terminal or no spawn in or if she doesn't need more help, she is removed from list.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,7 +8,6 @@ require('prototype_room_init');
 require('prototype_room_costmatrix');
 require('visualizer');
 require('screepsplus');
-require('visualizer');
 
 if (config.profiler.enabled) {
   try {
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -200,6 +200,74 @@ Room.prototype.handleScout = function() {
   }
 };
 
+Room.prototype.checkNeedHelp = function() {
+  let spawn = Game.rooms[this.name].find(STRUCTURE_SPAWN);
+  let needHelp = this.memory.energyAvailableSum < config.carryHelpers.needTreshold * config.carryHelpers.ticksUntilHelpCheck &&
+    !this.hostile && spawn && spawn.my;
+  let oldNeedHelp = this.memory.needHelp;
+  if (needHelp) {
+    if (!oldNeedHelp) {
+      this.memory.energyAvailableSum = 0;
+      Memory.needEnergyRooms.push(this.name);
+      this.memory.needHelp = true;
+      return '---!!!---' + this.name + ' need energy ---!!!---';
+    }
+    return 'Already set as needHelp';
+  }
+  if (oldNeedHelp) {
+    this.memory.energyAvailableSum = 0;
+    _.remove(Memory.needEnergyRooms, (r) => r === this.name);
+    delete Memory.rooms[this.name].needHelp;
+    return '---!!!---' + this.name + ' no more need help ---!!!---';
+  }
+  return;
+
+};
+Room.prototype.checkCanHelp = function(targetName) {
+  if (!Memory.needEnergyRooms) { return; }
+
+  let nearestRoom = this.memory.nearestRoom;
+  if (!nearestRoom || !Memory.rooms[nearestRoom].needHelp) {
+    nearestRoom = this.nearestRoomName(Memory.needEnergyRooms, config.carryHelpers.maxDistance);
+    this.memory.nearestRoom = nearestRoom;
+  }
+  if (!Game.rooms[nearestRoom] || !Memory.rooms[nearestRoom].needHelp) {
+    _.remove(Memory.needEnergyRooms, (r) => r === nearestRoom);
+  }
+
+  let canHelp = this.memory.energyAvailableSum > config.carryHelpers.helpTreshold * config.carryHelpers.ticksUntilHelpCheck &&
+    targetName !== this.name && Game.rooms[targetName] && this.storage &&
+    !Game.rooms[targetName].hostile && !this.terminal;
+  if (canHelp) {
+    this.checkRoleToSpawn('carry', config.carryHelpers.maxHelpersAmount, this.storage.id,
+      this.name, undefined, targetName);
+    this.memory.energyAvailableSum = 0;
+    return '---!!! ' + this.name + ' send energy to: ' + targetName + ' !!!---';
+  }
+  return 'no';
+
+};
+
+Room.prototype.checkForEnergyTransfer = function() {
+  Memory.needEnergyRooms = Memory.needEnergyRooms || [];
+  this.memory.energyAvailableSum = this.memory.energyAvailableSum || 0;
+
+  if (Game.time % config.carryHelpers.ticksUntilHelpCheck) {
+    let factor = config.carryHelpers.factor;
+    this.memory.energyAvailable = (1 - factor) * this.memory.energyAvailable + (factor) * this.energyAvailable || 0;
+    this.memory.energyAvailableSum += this.memory.energyAvailable;
+    return;
+  }
+  let needHelp = this.checkNeedHelp();
+  if (needHelp) {
+    if (needHelp !== 'Already set as needHelp') { this.log(needHelp); }
+  } else {
+    let canHelp = this.checkCanHelp();
+    if (canHelp !== 'no') { this.log(canHelp); }
+  }
+  this.memory.energyAvailableSum = 0;
+};
+
 Room.prototype.executeRoom = function() {
   this.buildBase();
 
@@ -326,6 +394,8 @@ Room.prototype.executeRoom = function() {
     }
   }
 
+  this.checkForEnergyTransfer();
+
   this.checkAndSpawnSourcer();
 
   if (this.controller.level >= 4 && this.storage) {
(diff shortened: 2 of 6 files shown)


Nevrdid:
https://github.com/TooAngel/screeps/pull/95

src/main.js
```diff
@@ -6,6 +6,7 @@ require('prototype_creep_move');
 require('prototype_roomPosition');
 require('prototype_room_init');
 require('prototype_room_costmatrix');
+
 require('visualizer');
 require('screepsplus');
 require('visualizer');
```
TooAngel: Why is this two times included?

Nevrdid: don't know ^^'

TooAngel: :-) I was not necessary meant as a question, more like a: 'Would you be so kind to remove one of the two?'

Thx, for fixing.

src/prototype_room_my.js
```diff
@@ -326,6 +326,57 @@ Room.prototype.executeRoom = function() {
     }
   }
 
+  let carryHelpInterval = config.carryHelpers.ticksUntilHelpCheck;
```
TooAngel: I think we should extract it to a method.

And I would initialize the use variables first, like;

```
Memory.needEnergyRooms = Memory.needEnergyRooms || [];
this.memory.energyAvailableSum = this.memory.energyAvailableSum || 0;
...
```

src/prototype_room_my.js
```diff
@@ -326,6 +326,57 @@ Room.prototype.executeRoom = function() {
...
+  let key = _.findKey(Memory.needEnergyRooms, this.name);
```
TooAngel: Move this to line 342 after `} else {`, no need to do this earlier

src/prototype_room_my.js
```diff
@@ -326,6 +326,57 @@ Room.prototype.executeRoom = function() {
...
+
+  if (Game.time % carryHelpInterval) {
+    this.memory.energyAvailableSum += this.energyAvailable;
```
TooAngel: You could work with a moving average: `this.memory.energyAvailable = (1 - factor) * this.memory.energyAvailable + (factor) this.energyAvailable`. E.g. factor 0.9 means, that the current value is very important, but take 0.1 from the previous moving average.
I think I would do something like 0.2. This way you have a better variable in memory and IMHO less complexity.

src/prototype_room_my.js
```diff
@@ -326,6 +326,57 @@ Room.prototype.executeRoom = function() {
...
+
+  } else if (Memory.needEnergyRooms === undefined) {
+    Memory.needEnergyRooms = [];
```
TooAngel: Maybe a collection? Easier to check for room existence and we could store something, maybe

src/prototype_room_my.js
```diff
@@ -326,6 +326,55 @@ Room.prototype.executeRoom = function() {
...
+    this.memory.energyAvailableSum += this.energyAvailable;
+
+  } else if (this.memory.energyAvailableSum === undefined) {
+    this.memory.energyAvailableSum = 0;
+
+  } else {
+    let needHelp = Memory.needEnergyRooms[key];
+    this.log(needHelp);
+    if (needHelp && this.hostile) {
+      Memory.needEnergyRooms[key] = undefined;
```
TooAngel: I usually do `delete Memory.needEnergyRooms[key]`

src/prototype_room_my.js
```diff
@@ -326,6 +326,55 @@ Room.prototype.executeRoom = function() {
...
+      Memory.needEnergyRooms = _.compact(Memory.needEnergyRooms);
```
TooAngel: So that you don't need this :-)

src/prototype_room_my.js
```diff
@@ -326,6 +326,55 @@ Room.prototype.executeRoom = function() {
...
+
+    } else if (this.memory.energyAvailableSum < config.carryHelpers.needTreshold * carryHelpInterval) {
+      if (!needHelp) {
+        Memory.needEnergyRooms.push(this.name);
+        this.log('---!!!---' + this.name + ' need energy ---!!!---');
+      }
```
TooAngel: Extracted to a method this could get much cleaner with `returrn`s and removing of all the nesting.

src/prototype_room_my.js
```diff
@@ -326,6 +326,55 @@ Room.prototype.executeRoom = function() {
...
+    } else if (needHelp) {
+      delete Memory.needEnergyRooms[this.name];
+    } else if (Memory.needEnergyRooms) {
+      let needHelp;
+      let nearestRoom = this.memory.nearestRoom;
```
TooAngel: We should add a ttl to the nearestRoom. E.g. when claiming a new room or loosing one this could change.

src/prototype_room_utils.js
```diff
@@ -8,6 +8,15 @@ Room.prototype.sortMyRoomsByLinearDistance = function(target) {
   return _.sortBy(Memory.myRooms, sortByLinearDistance);
 };
 
+Room.prototype.nearestRoomName = function(roomsNames, limit) {
+  let roomName = this.name;
+  let sortByLinearDistance = function(object) {
+    let dist = Game.map.getRoomLinearDistance(roomName, object);
+    return dist <= limit ? dist : 1000;
+  };
+  return _.sortBy(roomsNames, sortByLinearDistance)[0];
```
TooAngel: You can use `_.min(romsNames, sortByLinearDistance)`

src/prototype_room_utils.js
```diff
@@ -8,6 +8,15 @@ Room.prototype.sortMyRoomsByLinearDistance = function(target) {
   return _.sortBy(Memory.myRooms, sortByLinearDistance);
 };
 
+Room.prototype.nearestRoomName = function(roomsNames, limit) {
```
TooAngel: I think I wouldn't pass the `roomsNames` and just use `Memory.myRooms` in the method.

Also you can `_.filter` `myRooms` first, exclude `distance > limit` and the room itself.

src/prototype_room_my.js
```diff
@@ -326,6 +326,55 @@ Room.prototype.executeRoom = function() {
...
+      if ((nearestRoom !== this.name) && this.memory.energyAvailableSum > config.carryHelpers.helpTreshold * carryHelpInterval &&
+        this.storage && nearestRoom) {
+        if (nearestRoom && Game.rooms[nearestRoom] && !Game.rooms[nearestRoom].hostile) {
+          this.checkRoleToSpawn('carry', config.carryHelpers.maxHelpersAmount, this.storage.id,
+            this.name, undefined, nearestRoom);
+          this.log('---!!! ' + this.name + ' send energy to: ' + nearestRoom + ' !!!---');
+        }
+      } else {
+        Memory.needEnergyRooms[key] = undefined;
+        Memory.needEnergyRooms = _.compact(Memory.needEnergyRooms);
```
TooAngel: `delete` and no `_.compact`, like above.

src/prototype_room_utils.js
Nevrdid: no because roomsNames come frome needEnergyRooms. i could use it inside but it still allow us to use it with myRooms for example somewhere else  :)

src/prototype_room_my.js
Nevrdid: actually a room stop to need help herself when she go up to the treshold for need help or if hostile exist. Else, i just added that an helper will remove his target if she no more need help or no more in Game.rooms. For new rooms, i added a check for see if there is a spawn in room.

src/prototype_room_my.js
Nevrdid: dunno, finally got a pretty easy way to manage it and i alway see arrays as 'light' collections :)

src/prototype_room_my.js
Nevrdid: _.remove :)

src/prototype_room_my.js
Nevrdid: let's try

Nevrdid:
I see it as mergeable :)

src/prototype_room_my.js
```diff
@@ -200,6 +200,49 @@ Room.prototype.handleScout = function() {
   }
 };
 
+Room.prototype.checkForEnergyTransfer = function() {
+  let carryHelpInterval = config.carryHelpers.ticksUntilHelpCheck;
+  if (Game.time % carryHelpInterval) {
```
Nevrdid: yep, it must. Game.time%interval check is true each time exept each (config value) ticks

src/prototype_room_my.js
```diff
@@ -200,6 +200,49 @@ Room.prototype.handleScout = function() {
...
+
+    let factor = config.carryHelpers.factor;
+    this.memory.energyAvailable = (1 - factor) * this.memory.energyAvailable + (factor) * this.energyAvailable;
+    this.memory.energyAvailableSum += this.memory.energyAvailable;
+    return;
+  }
+
+  Memory.needEnergyRooms = Memory.needEnergyRooms || [];
+  this.memory.energyAvailableSum = this.memory.energyAvailableSum || 0;
```
Nevrdid: don't really need to check that each ticks. only when check launched

src/prototype_room_my.js
```diff
@@ -200,6 +200,49 @@ Room.prototype.handleScout = function() {
...
+  let needHelp = this.memory.needHelp;
+  if (!needHelp && this.memory.energyAvailableSum < config.carryHelpers.needTreshold * carryHelpInterval && !this.hostile && Game.rooms[this.name].find(STRUCTURE_SPAWN)) {
+    this.memory.energyAvailableSum = 0;
+    Memory.needEnergyRooms.push(this.name);
+    this.memory.needHelp = true;
+    return '---!!!---' + this.name + ' need energy ---!!!---';
+  } else if (needHelp) {
+    _.remove(Memory.needEnergyRooms, (r) => r === this.name);
+    delete Memory.rooms[this.name].needHelp;
```
Nevrdid: yeah because the if check first if it need help. so if it come here, it don't need help

TooAngel:
Better, fine with me, could you squash the commits please?

PR #99, opened 2017-01-07
[WIP] Fix Nextroomer 2nd try

Tsungen:
- [ ] add check when to fill energy and spawn sourcer
- [ ] change trigger for ative

### DIFF
--- a/src/role_nextroomer.js
+++ b/src/role_nextroomer.js
@@ -308,11 +308,15 @@ roles.nextroomer.settle = function(creep) {
     methods.push(Creep.constructTask);
   }
 
-  if (creep.room.controller.level < 8) {
-    methods.push(Creep.upgradeControllerTask);
+  // Fill energy after building is done
+  if (creep.room.controller.level >= 4) {
+    methods.push(Creep.transferEnergy);
   }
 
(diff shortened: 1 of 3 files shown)


src/role_nextroomer.js
```diff
@@ -308,11 +308,15 @@ roles.nextroomer.settle = function(creep) {
     methods.push(Creep.constructTask);
   }
 
+  // Fill energy after building is done
+  if (creep.room.controller.level >= 4) {
+    methods.push(Creep.transferEnergy);
+  }
+
   if (creep.room.controller.level < 8) {
+
```
Somotaw: Is this really what we want? Personally i'd prefer if they used their valuable work parts for upgrading.

Tsungen: As far as i understand the nextroomer role.
Build up a new room to a certain level or revive/rebuild a fallen room

with the other checks they will upgrade less after RCL4 i reached. maybe we can remove the check.

Nevrdid: the problem is the lack of energy. try https://github.com/TooAngel/screeps/pull/98 i think it could resolve nextroom which harvest to slowly and build roads by the same.

PR #123, opened 2017-01-10
[WIP] hotfix/Fixes to squad pathing

Somotaw:
Also issue with undefined variable in squadheal.

Tried this and it seems to work, get back to me with issues :)


Somotaw:
hmm this might be the real culprit, but my squadhealers bug out if room to be attacked is 1 tile away, mayeb add else after and check for if movetarget = baseroom

PR #127, opened 2017-01-10
hotfix/carry-range-check #125

### LINKED ISSUE #125: Carry Issue
carrys went back and forth in new room

changed carry line 41 to
if (creep.pos.getRangeTo(Game.getObjectById(targetId)) <= 1) {
seems fixed

Somotaw:
This change can be introduced here too to resolve #125 ?

### DIFF
--- a/src/role_carry.js
+++ b/src/role_carry.js
@@ -38,7 +38,7 @@ roles.carry.preMove = function(creep, directions) {
       // TODO better the position from the room memory
       if (source !== null) {
         let returnCode = creep.moveTo(Game.getObjectById(targetId));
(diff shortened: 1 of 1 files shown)


src/role_carry.js
```diff
@@ -38,7 +38,7 @@ roles.carry.preMove = function(creep, directions) {
       // TODO better the position from the room memory
       if (source !== null) {
         let returnCode = creep.moveTo(Game.getObjectById(targetId));
-        if (creep.pos.getRangeTo(Game.getObjectById(targetId)) > 1) {
```
TooAngel: Stay as it is

PR #133, opened 2017-01-12
Fix scout stuck issue

Somotaw:
This works, but maybe not the right way to fix it? Comments welcome.

### DIFF
--- a/src/role_scout.js
+++ b/src/role_scout.js
@@ -115,11 +115,11 @@ roles.scout.execute = function(creep) {
           delete creep.memory.last; // Delete to reset stuckness.
         }
       }
-      if (search.path.length > 0) {
-        creep.move(creep.pos.getDirectionTo(search.path[0]));
-      } else {
-        creep.moveTo(targetPosObject);
-      }
+      //if (search.path.length > 0) {
+      //creep.move(creep.pos.getDirectionTo(search.path[0]));
+      //} else {
+      creep.moveTo(targetPosObject);
+      //}
       return true;
     }
     creep.say(creep.pos.getDirectionTo(search.path[0]));


TooAngel:
DCBOT: This repository is under [democratic collaboration](https://github.com/TooAngel/democratic-collaboration) and will be merged automatically.

Somotaw:
Nice bot is talking \o/

src/role_scout.js
```diff
@@ -115,11 +115,11 @@ roles.scout.execute = function(creep) {
           delete creep.memory.last; // Delete to reset stuckness.
         }
       }
-      if (search.path.length > 0) {
```
Nevrdid:
```
if (search.path.length === 0 || (search.incomplete && !search[1])) {
  creep.moveTo(targetPosObject);
} else {
  creep.move(creep.pos.getDirectionTo(search.path[0]));
}
```

src/role_scout.js
```diff
@@ -115,11 +115,11 @@ roles.scout.execute = function(creep) {
...
-        creep.move(creep.pos.getDirectionTo(search.path[0]));
-      } else {
-        creep.moveTo(targetPosObject);
-      }
+      //if (search.path.length > 0) {
+      //creep.move(creep.pos.getDirectionTo(search.path[0]));
+      //} else {
+      creep.moveTo(targetPosObject);
+      //}
```
Nevrdid: if (search.path.length === 0 || (search.incomplete && !search[1])) {
creep.moveTo(targetPosObject);
} else {
creep.move(creep.pos.getDirectionTo(search.path[0]));
}

TooAngel:
DCBOT: Code update recognized, countdown starts fresh.

TooAngel:
DCBOT: Code update recognized, countdown starts fresh.

TooAngel:
I don't have a good solution for this.

My opinion on this. I think a `PathFinder` with `range > 0` config returns always `incomplete`. Not tested well but I run into something like that a couple of times.
So the proper fix should start there, they code you are editing is only the backup case. Maybe there is also a better solution for the actual move, instead of `RoomPosition(25, 25)`

TooAngel:

Somotaw:
incomplete - If the pathfinder fails to find a complete path, this will be true. Note that `path` will still be populated with a partial path which represents the closest path it could find given the search parameters.

So we should remove if path incomplete from the if?

TooAngel:
> incomplete - If the pathfinder fails to find a complete path, this will be true. Note that path will still be populated with a partial path which represents the closest path it could find given the search parameters.

If I configure the PathFinder with a range, will it always say incomplete. Because the path is not calculated until the end? I got the impression in this case.

Somotaw:
range number
Range to pos before goal is considered reached. The default is 0.

So if at range 20 goal should be considered reached ok. but that means that if its tryong to move OUT of a room it will consider path incomplete when getting close to the edge :)

Somotaw:
somehow related to code changing the memory.search.target I'm guessing

TooAngel:
So when I working on it. I got the impression that the scout will always run into the backup case and not use the PathFinder properly. We can do your quick fix, I'm fine with it. But I think the PathFinder should work proper and we really have to address only the special cases.

Somotaw:
I dont want to do the quick and easy fix, tryign to find culprit :D

TooAngel:
DCBOT: Code update recognized, countdown starts fresh.

TooAngel:
DCBOT: Code update recognized, countdown starts fresh.

TooAngel:
DCBOT: Code update recognized, countdown starts fresh.

TooAngel:
DCBOT: Code update recognized, countdown starts fresh.

src/role_scout.js
Somotaw: Why do we want it to use the path is the pat his incomplete, the built path will be ov no use it it doesnt reach all the way?
 if (search.path.length > 0)

 seems of no use to us when path is inclomplete here 

if (search.incomplete || search.path.length === 0 || (creep.inBase() && creep.room.memory.misplacedSpawn)) {

Somotaw: since pathfinder will return a path even though it cannor reach target path length will almost always be above 0

Somotaw: So we cant chekc for path length, we have to check for last path position = target?

TooAngel:
DCBOT: Removing WIP label, the title is not prefixed with [WIP]

TooAngel:
DCBOT: Removing WIP label, the title is not prefixed with [WIP]

TooAngel:
DCBOT: A new review, yeah.

    Votes: 218/287
    Coefficient: 0.759581881533
    Hours to merge: -13
    Age in hours: 22

PR #175, opened 2017-01-25
[WIP] Fix scout stuck issue

Somotaw:
This is technically an issue with the new costmatrix branch as ignore creeps is set to true instead of false as method and the proper check is commented out + wrong, which I fix here, need to make sure to merge this correctly when applying costmatrixbranch.

We should have this even wtihout the costmatrix cause this is how we get it WAI

### DIFF
--- a/src/role_scout.js
+++ b/src/role_scout.js
@@ -91,7 +91,7 @@ roles.scout.execute = function(creep) {
       search = PathFinder.search(
         creep.pos, {
           pos: targetPosObject,
-          range: 20
+          range: 25
         }, {
           roomCallback: creep.room.getAvoids(creep.room, {
             pos: targetPosObject,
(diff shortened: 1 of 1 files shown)


TooAngel:
DCBOT: 
This repository is under [democratic collaboration](https://github.com/TooAngel/democratic-collaboration) and will be merged automatically.

The merge decision is based on the outcome of the reviews:
 - `Approve` add the reviewer value (number of commits) to the `metric`
 - `Request changes` substract the reviewer value from the `metric`

 The merge will happen after `(1 - metric/total.votes) * (5 + pull_request.commits * 5))` calculated from the last code change.

Please review the PR to make a good democratic decision.

Summoning some reviewers:
 - @TooAngel: Some warm words would be nice
 - @Nevrdid: Please look at the code
 - @WhiteHalmos: Not sure why you are here

Somotaw:
In this case its the other way around, the scout gets blocked by for example sourcers mining on path and then blocks other creeps behind it

Somotaw:
also it only uses this when search is incomplete

Somotaw:
fixed that part so

Somotaw:
image
this room has the issue, sroucer at bottom left blocks scout and therefore carrys etc etc

TooAngel:
My idea was to use `getAvoid` or upcoming `moveToMy` which will avoid calculated creep / structure positions.

Best would be solving the underlying issue, so that we don't need the fallback.

It was also discussed if we just use the usual routing, and overwrite `memory.base` each time we use a new route.

Somotaw:
I like your idea :)

democratic-collaboration:

PR #179, opened 2017-01-25
Build terminal and labs next to extractor path.

Somotaw:


TooAngel:
DCBOT: 
This repository is under [democratic collaboration](https://github.com/TooAngel/democratic-collaboration) and will be merged automatically.

The merge decision is based on the outcome of the reviews:
 - `Approve` add the reviewer value (number of commits) to the `metric`
 - `Request changes` substract the reviewer value from the `metric`

 The merge will happen after `(1 - metric/total.votes) * (5 + pull_request.commits * 5))` calculated from the last code change.

Please review the PR to make a good democratic decision.

Summoning some reviewers:
 - @TooAngel: I just want to get this merged ASAP
 - @Nevrdid: Some warm words would be nice
 - @WhiteHalmos: A proper review is appreciated

Somotaw:
image

democratic-collaboration:
DCBOT: A new review, yeah.

    Votes: 78/302
    Coefficient: 0.258278145695
    Merging in 6 days 3 hours
    Age 1 days 7 hours

src/prototype_room_init.js
```diff
@@ -286,6 +310,8 @@ Room.prototype.setup = function() {
   });
   let paths_sorted = _.sortBy(paths_controller, sorter);
   let path = this.getMemoryPath(paths_sorted[paths_sorted.length - 1].name);
+  let pathLB = this.getMemoryPath(paths_controller[4].name);
```
TooAngel: Why `[4]`.

You can get the mineral path, by getting the id and use `pathStart-ID` path.

src/prototype_room_init.js
```diff
@@ -97,11 +133,6 @@ function setStructures(room, path, costMatrixBase) {
         costMatrixBase.set(structurePos.x, structurePos.y, config.layout.structureAvoid);
         continue;
       }
-      if (room.memory.position.structure.lab.length < CONTROLLER_STRUCTURES.lab[8]) {
-        room.memory.position.structure.lab.push(structurePos);
-        costMatrixBase.set(structurePos.x, structurePos.y, config.layout.structureAvoid);
-        continue;
-      }
```
TooAngel: I'm tending to keep this. You set the structures before, so nothing should happen here.

If the mineral path is too short or whatever, it would place the rest of the labs and terminal here.

src/prototype_room_init.js
```diff
@@ -63,6 +63,42 @@ Room.prototype.setTowerFiller = function() {
...
+  for (let pathI = path.length - 1; pathI > 0; pathI--) {
+    let pathPos = new RoomPosition(path[pathI].x, path[pathI].y, room.name);
+    let structurePosIterator = pathPos.findNearPosition();
+    for (let structurePos of structurePosIterator) {
+      if (room.memory.position.structure.lab.length < CONTROLLER_STRUCTURES.lab[8]) {
+        room.memory.position.structure.lab.push(structurePos);
+        costMatrixBase.set(structurePos.x, structurePos.y, config.layout.structureAvoid);
+        continue;
+      }
+      if (room.memory.position.structure.terminal.length < CONTROLLER_STRUCTURES.terminal[8]) {
```
TooAngel: Set terminal first, so that it is close to the `extractor` creep

src/prototype_room_init.js
```diff
@@ -63,6 +63,42 @@ Room.prototype.setTowerFiller = function() {
...
+      if (room.memory.position.structure.lab.length < CONTROLLER_STRUCTURES.lab[8] ||
+        room.memory.position.structure.terminal.length < CONTROLLER_STRUCTURES.terminal[8]) {
+        room.log('Structures not found: ' +
+          'lab: ' + room.memory.position.structure.lab.length + ' ' +
+          'terminal: ' + room.memory.position.structure.terminal.length
+        );
+        continue;
+      }
+      if (!room.memory.position.pathEnd) {
+        room.log('Room not completly build');
```
TooAngel: You will always get this error, because `setStructures` is called after this and `pathEnd` is set there

src/prototype_room_init.js
Somotaw: That would use more cpu wouln't it? we always generate path's in same order no?

src/prototype_room_init.js
Somotaw: hmm maybe, would need testing in rooms where space is cramped

src/prototype_room_init.js
Somotaw: I would rather have it closer to storage?

src/prototype_room_init.js
Somotaw: right, removing.

src/prototype_room_init.js
TooAngel: If you have a `!= 2` sources room it will not fit.

I tried to stay independent from as much as possible.

Somotaw: ahh thought we didn't settle 1 source rooms but I see what you mean.

src/prototype_room_init.js
TooAngel: I think it is better to get the mineral out of the extractor faster. If the terminal is close, the extractor creep would be more efficient.

Distribution can be optimized with more or other creeps.

src/prototype_room_init.js
TooAngel: If the path is not long enough if will be ignored. As seen in current rooms where the labs are missing.

src/prototype_room_init.js
TooAngel: we shouldn't. But I would also like to support different private server mods

PR #54, opened 2016-12-29
CreepBuilder improvments

Nevrdid:
Actually, all looks to work well :)

PR #188, opened 2017-02-03
[WIP] Added Alternative Spawn when Queue length > 5 or TTL expired.

invisioblack:
Thought process is - whatever is generated is better than having a stalled queue or items timing out.

### DIFF
--- a/src/prototype_room_creepbuilder.js
+++ b/src/prototype_room_creepbuilder.js
@@ -117,52 +117,51 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
 
     let creep = this.memory.queue[0];
     energyNeeded = 50;
-    let c_room = this;
+    let c_room = this;    
     let sortByDistance = function(object) {
       return Game.map.getRoomLinearDistance(c_room.name, object);
     };
-    let roomsMy = _.sortBy(Memory.myRooms, sortByDistance);
+    let roomsMy = _.sortBy(Memory.myRooms, sortByDistance);
     if (this.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, this.name)) {
       console.log('['+this.name +' Spawning]- Activating '+creep.role );
-      this.memory.queue.shift();    
-    } else {
-      if (creep.ttl === 0) {
+      this.memory.queue.shift();
+    } else if (creep.ttl) {
+      if ( creep.ttl === 0) {
         for (var room_id in roomsMy) {
           var s_room = Game.rooms[roomsMy[room_id]];
           if (s_room !== this) {
             if (s_room.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, this.name)) {
               console.log('['+s_room.name +' Spawning]- Activating '+creep.role+' for: '+this.name  );
-              this.memory.queue.shift;      
-              return;     
-            }                 
-          }   
-        }  
+              this.memory.queue.shift();
+              return;
+            }
+          }           
+        }
+        this.log('TTL reached, skipping: ' + JSON.stringify(creep));
+        this.memory.queue.shift();
+        return;
       }
-      this.log('TTL reached, skipping: ' + JSON.stringify(creep));
-      this.memory.queue.shift();
-      return;
-    }      
+    }; 
     // TODO maybe skip only if there is a spawn which is not spawning
     creep.ttl = creep.ttl || config.creep.queueTtl;
     let spawnsNotSpawning = _.filter(this.find(FIND_MY_SPAWNS), function(object) {
-      return !object.spawning;      
-    });      
+      return !object.spawning;
+    });
     if (spawnsNotSpawning.length === 0) {
-      creep.ttl--;      
-    }    
-  } 
-  if(this.memory.queue.length > 5){
-    for (var room_id in roomsMy) {
-      var s_room = Game.rooms[roomsMy[room_id]];
-      if (s_room !== this) {
-        if (s_room.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, this.name)) {
-          console.log('['+s_room.name +' Spawning]- Activating '+creep.role+' for: '+this.name  );
-          this.memory.queue.shift;
-          return;     
-        }                 
-      }   
-    } 
-  }
+      creep.ttl--;
+    }
+    if(this.memory.queue.length > 5){
+      for (var room_id in roomsMy) {
+        var s_room = Game.rooms[roomsMy[room_id]];
+        if (s_room !== this) {
+          if (s_room.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, this.name)) {
+            console.log('['+s_room.name +' Spawning]- Activating '+creep.role+' for: '+this.name  );
+            this.memory.queue.shift();
+            return;
+          }
+        }
+      }
+    }
     // Spawing only one per tick
     return;
   }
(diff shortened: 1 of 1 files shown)


democratic-collaboration:

democratic-collaboration:
Every approved review will speed up the merge, request changes will slow it down.

Please review the PR to help.

invisioblack:
Testing Result- is de-queuing for TTL expiration too soon.  Misplaced ;.  Validating on private server for GCL 3.  Because testing is limited to 2 rooms - Validation with more rooms is needed.

TO DO - 
Configuration to turn on and off

TooAngel:
Didn't had a look at the implemenation yet, hopefully in a couple of hours.

Some creeps have a `stayInRoom` flag set, which makes sure they don't leave the base, maybe this could make problems / needs to be improved that it moves back to base.

invisioblack:
Actually - stay in room is what makes this work. You assign the room that had queued the item as the base rather than where the item is spawning- so it detects it is out of it's room and tries to move back.  That functionality works really well.

src/prototype_room_creepbuilder.js
```diff
@@ -117,23 +117,49 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
 
     let creep = this.memory.queue[0];
     energyNeeded = 50;
-
-    if (this.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, creep.base)) {
+    let c_room = this;
```
TooAngel: In the rest of the code we just use `room`

src/prototype_room_creepbuilder.js
```diff
@@ -117,23 +117,49 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
...
+    };
+    let roomsMy = _.sortBy(Memory.myRooms, sortByDistance);
+    if (this.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, this.name)) {
+      console.log('['+this.name +' Spawning]- Activating '+creep.role );
       this.memory.queue.shift();
-    } else {
-      if (creep.ttl === 0) {
+    } else if (creep.ttl) {
+      if ( creep.ttl === 0) {
+        for (var room_id in roomsMy) {
```
TooAngel: Initialize the `roomsMy` directly before the `for` loop to save CPU.

src/prototype_room_creepbuilder.js
```diff
@@ -117,23 +117,49 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
...
+          var s_room = Game.rooms[roomsMy[room_id]];
+          if (s_room !== this) {
```
TooAngel: You should filter the rooms, also for rooms too far away. Many of my rooms can not reach some other rooms. In this step also remove `this` room. And I would do that before ordering.

src/prototype_room_creepbuilder.js
```diff
@@ -117,23 +117,49 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
...
+            if (s_room.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, this.name)) {
+              console.log('['+s_room.name +' Spawning]- Activating '+creep.role+' for: '+this.name  );
```
TooAngel: use `s_room.log` or `this.log` to use the usual logging prefix.

src/prototype_room_creepbuilder.js
```diff
@@ -117,23 +117,49 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
...
+    }; 
+    // TODO maybe skip only if there is a spawn which is not spawning
+    creep.ttl = creep.ttl || config.creep.queueTtl;
+    let spawnsNotSpawning = _.filter(this.find(FIND_MY_SPAWNS), function(object) {
+      return !object.spawning;
+    });
+    if (spawnsNotSpawning.length === 0) {
+      creep.ttl--;
+    }
+    if(this.memory.queue.length > 5){
```
TooAngel: Extract the logic to some method, so that you don't duplicate it.

src/prototype_room_creepbuilder.js
```diff
@@ -117,23 +117,49 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
...
+      for (var room_id in roomsMy) {
```
TooAngel: You can use `for (let roomName of roomsMy) {let s_room = Game.rooms[roomName];}

src/prototype_room_creepbuilder.js
invisioblack: Removed this line as it is unneeded. This is aliased to room at the top of the if block.  Also modifying the rest of the file to follow the correct idiom

invisioblack: Removed this line as it is unneeded. This is aliased to room at the top of the if block.  Also modifying the rest of the file to follow the correct idiom

src/prototype_room_creepbuilder.js
```diff
@@ -67,9 +67,9 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
   let storages;
   let energyNeeded;
   let unit;
-
-  if (this.memory.queue.length > 0 && (creepsConfig.length === 0 || creepsConfig[0] != 'harvester')) {
-    let room = this;
+  let room = this;
```
invisioblack: Moved to outside for loop so that room is available in the entire function

src/prototype_room_creepbuilder.js
```diff
@@ -113,36 +113,53 @@ Room.prototype.spawnCheckForCreate = function(creepsConfig) {
...
+      if (dist < 10 && dist > 0){
+        return dist;    
+      }      
+      return;
+    };
+   
+    if (room.spawnCreateCreep(creep.role, creep.heal, creep.level, creep.squad, creep.routing, room.name)) {
+      room.log('['+room.name +' Spawning]- Activating '+creep.role );
+      room.memory.queue.shift();
+    } else if ((creep.ttl && creep.ttl === 0)|| room.memory.queue.length > 5) {
```
invisioblack: Condensed into one for loop that covers both cases.

invisioblack: condensed to one for loop that covers both cases.

TooAngel:
Squashing and we should be fine

src/prototype_room_creepbuilder.js
Nevrdid: this is already available for entire function :) i personally prefer to use this when we can.
for use in others functions like sortByDistance you can move them as room method and so aply them to this ( plus, i don't remember why but i believe this method already exist somewhere).

Nevrdid:
I sugest you to see what i did here : https://github.com/TooAngel/screeps/pull/54/files#diff-72fe2da08c0bdf7fdd904d0252f0cd72 as it will avoid me some conflicts when it will be merged ^^'

invisioblack:
Issues found in testing - 
1. Creeps with 'stay in room: false' attempt to use pathing.  If there is no pathing for any reason - they error out.  Newbie highway rooms are a great example

2. Harvesters are queued in new rooms - even before the spawn is created.  this is not a problem unless the room also has a hostile 'visitor' when the defence code can also queue defenders. This puts the queue over size and causes harvesters to build every tick until some get to wherever the room is. 

Possible solutions:
1. In addtion to the configuration option to turn on and off add a manual flag stored ont he room to turn on and off. Set ' eligible to help and eligible to receive help levels.  Perhaps require that the room have a spawn? Restrict the number of 'Assistant Rooms- require that the rooms asissting not reflect so that is room A being helped by room B would not help room B back.

2. Use a 'crash test dummy' to check the path to the rooms to be helped.  Time how long it takes.  if it takes over 500 ticks - report the path as too long(or failed).  Possibly retest every 10000 ticks?

invisioblack:
Nextroomers & 218 handles this better. will rarely run.

PR #213, opened 2017-02-23
Fix grunt screeps script folder

Nevrdid:
This is a right fix for sure but i still have an error 401 when i try to upload. Dunno where it come from.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,7 +33,7 @@ module.exports = function(grunt) {
         ptr: false
       },
       dist: {
-        src: ['dist/*.js']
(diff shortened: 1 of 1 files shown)


TooAngel:
You are sure you set your credentails correctly?

TooAngel:
In option would also be:

Define multiple targets for the `screeps` task in grunt.
I'm fine to use the `src/` folder as the default target for `grunt screeps`, but we should habe a second one, which deploys the `dist/` folder and is used in the default grunt task, executed by just `grunt`.

Nevrdid:
ok, i'll try to do it dunno when :)

PR #218, opened 2017-03-04
Handle Stuck &&  Harvester Fill Misplaced Spawn for RCL 4

invisioblack:
This has 4 items in it -
1. isStuck updated to store last stuck position - also tracks 5 positions instead of 3.

2. Handle Stuck for Carry, Harvester, Planer & repairer & scout.  They should move every tick, if they have not moved for 105 ticks something is really wrong.

3. Harvester filling spawn when the spawn is misplaced (not on path) at RCL 4

### DIFF
--- a/src/prototype_creep.js
+++ b/src/prototype_creep.js
@@ -84,7 +84,7 @@ Creep.prototype.handle = function() {
     if (this.memory.last === undefined) {
       this.memory.last = {};
     }
-    if(this.fatigue === 0){
+    if (this.fatigue === 0){
       let last = this.memory.last;
       this.memory.last = {
         pos1: this.pos,
@@ -98,17 +98,17 @@ Creep.prototype.handle = function() {
   }
 };
 Creep.prototype.isStuck = function() {
-  if(this.memory.last !== undefined && this.memory.last.pos5 !== undefined && this.pos.isEqualTo(this.memory.last.pos5.x, this.memory.last.pos5.y)){
-    if(!this.memory.stuckCount){
+  if (this.memory.last !== undefined && this.memory.last.pos5 !== undefined && this.pos.isEqualTo(this.memory.last.pos5.x, this.memory.last.pos5.y)) {
+    if (!this.memory.stuckCount) {
       this.memory.stuckCount = 0;
     }
-    if(!this.memory.stuckx){
+    if (!this.memory.stuckx) {
       this.memory.stuckx = this.memory.last.pos1.x;
     }
-    if(!this.memory.stucky){
+    if (!this.memory.stucky) {
       this.memory.stucky = this.memory.last.pos1.y;
     }
-    if(this.memory.stuckx == this.memory.last.pos1.x && this.memory.stucky == this.memory.last.pos1.y){
+    if (this.memory.stuckx == this.memory.last.pos1.x && this.memory.stucky == this.memory.last.pos1.y) {
       return true;
     } else {
       return false;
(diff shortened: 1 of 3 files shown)


democratic-collaboration:

Please review the PR to help.

Summoning some reviewers:
 - @TooAngel: Some warm words would be nice
 - @Nevrdid: A proper review is appreciated
 - @WhiteHalmos: Who are you?

src/prototype_creep.js
```diff
@@ -84,21 +84,50 @@ Creep.prototype.handle = function() {
...
-      pos3: last.pos2,
-    };
+    if(this.fatigue === 0){
+      let last = this.memory.last;
+      this.memory.last = {
+        pos1: this.pos,
+        pos2: last.pos1,
+        pos3: last.pos2,
+        pos4: last.pos3,
+        pos5: last.pos4,
```
TooAngel: Wouldn't it be easier to store a list instead of the single positions?

src/prototype_creep.js
```diff
@@ -84,21 +84,50 @@ Creep.prototype.handle = function() {
...
-
 Creep.prototype.isStuck = function() {
-  return this.memory.last !== undefined &&
-    this.memory.last.pos3 !== undefined &&
-    this.pos.isEqualTo(this.memory.last.pos3.x, this.memory.last.pos3.y);
+  if(this.memory.last !== undefined && this.memory.last.pos5 !== undefined && this.pos.isEqualTo(this.memory.last.pos5.x, this.memory.last.pos5.y)){
+    if(!this.memory.stuckCount){
+      this.memory.stuckCount = 0;
+    }  if(!this.memory.stuckx){
+      this.memory.stuckx = this.memory.last.pos1.x;
```
TooAngel: I would use an object `this.memory.stuck = {x: 0, y:0}`

src/prototype_room_my.js
```diff
@@ -432,10 +432,26 @@ Room.prototype.executeRoom = function() {
...
+  if(harvesters.length < 1){
     this.spawnCheckForCreate();
   }
-
+  if(this.memory.spawnTimer > 100 ||  this.energyAvailable == this.energyCapacityAvailable || (this.controller.level *250 < this.energyCapacityAvailable && Game.Time % 50 == 0 ) || this.memory.attackTimer > 15){
+    this.memory.spawnTimer = 0;
+    this.spawnCheckForCreate();
+  } else {
+    this.memory.spawnTimer = this.memory.spawnTimer + 1;
+  }
```
TooAngel: What is your reasoning behind this block of code?

src/prototype_creep.js
invisioblack: It might be - I just modified what was already there by adding two more.

src/prototype_room_my.js
invisioblack: it causes the spawn to run 100 ticks after the last time it ran.  The goal is for it to always run when full - but of we use a mod Game.time for the last fallback it could happen in the middle of a fill cycle. this way it will run in the event of a room failure - but it will not run randomly.  It is the ultimate failure case.

src/prototype_creep.js
```diff
@@ -84,21 +84,53 @@ Creep.prototype.handle = function() {
...
+  } else {
+    return false;
+  };
+};
+Creep.prototype.handleStuck = function() {
+  if(this.isStuck() && (this.role == 'harvester' || this.role == 'carry' || this.role == 'planer' || this.role == 'repairer' || this.role == 'scout')){
+    this.memory.stuckCount = this.memory.stuckCount + 1;
+    this.log('Stuck! - ' + this.memory.stuckCount);
+    if(this.memory.stuckCount > 100){
+      this.log('UnStuck!')
```
TooAngel: Make the message more prominent and maybe with some debug output. We shouldn't come to this situation.

src/prototype_room_my.js
TooAngel: I don't think this is a good idea, the spawn should be constantly spawning.
The queue should make sure that the most important creep will be spawned.
The creeps should make sure that they are spawned in the best size possible.

So if your problem is that, creeps are spawned which could have been bigger, if they are spawned later. The logic for that, should live in the creep config logic.

invisioblack: Ok.

PR #242, opened 2017-03-31
Room.executeEveryTick

TooAngel:
Extract method to check if a certain needs to be executed on this tick.
To make sure the execution of the same task in different rooms, the
position of the controller is used as offset to the current Game tick.


democratic-collaboration:

Please review the PR to help.

PR #251, opened 2017-06-02
Do not care about hostile creeps if safe mode is active

samogot:
Do not care about hostile creeps if safe mode is active (and will not end until all hostile creeps are die)

### DIFF
--- a/src/role_nextroomer.js
+++ b/src/role_nextroomer.js
@@ -200,7 +200,7 @@ roles.nextroomer.underSiege = function(creep) {
 
 roles.nextroomer.settle = function(creep) {
   let room = Game.rooms[creep.room.name];
-  let hostileCreeps = room.find(FIND_HOSTILE_CREEPS);
+  let hostileCreeps = room.find(FIND_HOSTILE_CREEPS, {filter: creep => !room.controller.safeMode || creep.ticksToLive > room.controller.safeMode});
   if (hostileCreeps.length) {
     room.memory.underSiege = true;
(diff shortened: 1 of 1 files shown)


democratic-collaboration:

Please review the PR to help.

democratic-collaboration:
A new review, yeah.

    Votes: 0/358
    Coefficient: 0.0
    Merging in 9 days 22 hours
    Age 0 days 1 hours

democratic-collaboration:
A new review, yeah.

    Votes: 181/358
    Coefficient: 0.505586592179
    Merging in 4 days 21 hours
    Age 0 days 1 hours

samogot:
@TooAngel The logic now it to keep the `underSiege` on false if there is no hostile creeps that will be alive when safemode ends. If at any time one will appear, `underSiege` will be turned on. But if for example the only Invader will die in 500 ticks - we should not spent 1000 ticks on preparation to nothing.

TooAngel:
If the idea is mainly for invaders, we could exclude these.

The `underSiege` should instruct `nextroomers` to build ramparts and towers next to the sources, to be able to defend. If these are build they **should** build the base as if it is not `underSiege`.

Still I'm fine with the current changes.

samogot:
> The underSiege should instruct nextroomers to build ramparts and towers next to the sources, to be able to defend. If these are build they should build the base as if it is not underSiege.

@TooAngel oh I see. But then there is another bug - this isn't work if tower can't be build due to RCL. If RCL is't enough `underSiege` is deleted, but on the next tick it is again set to true - so no base building is done

PR #252, opened 2017-06-02
Spawn creeps beforehand

samogot:
Spawn creeps beforehand. Do not count creeps close to death, so that new creep will be ready exactly when old one will die (if energy is enough)

### DIFF
--- a/src/prototype_room_creepbuilder.js
+++ b/src/prototype_room_creepbuilder.js
@@ -84,6 +84,8 @@ Room.prototype.inRoom = function(creepMemory, amount = 1) {
   this.memory.roles = this.memory.roles || {};
 
   let j = 0;
+  const bodyPartConfig = this.getPartConfig(creepMemory, true);
+  const ticksToSpawn = bodyPartConfig ? bodyPartConfig.length * CREEP_SPAWN_TIME : null;
   for (let creep of creeps) {
     let iMem = Memory.creeps[creep.name];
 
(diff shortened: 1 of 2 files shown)


democratic-collaboration:

Please review the PR to help.

samogot:
I don't analyse spawn and queuing logic deep enough, and i think there can be a lot of improvements in this direction, but that change will give some improvement already now

TooAngel:
You should / could have a look at https://github.com/TooAngel/screeps/blob/527f2267235779c504f4feeb6436f07c602f2191/src/prototype_creep.js#L329

When a creep (sourcer, upgrader, ...) starts to work it stores the current ttl and instructs the base room to queue a replacement for that time. The move time should be handled this way, I'm not sure if the spawn time is properly included.

Anyway, wanted to take some time to check your PR and the implications. Will do that hopefully later today.

src/prototype_room_creepbuilder.js
```diff
@@ -285,7 +288,7 @@ Room.prototype.getPartConfig = function(creep) {
   if (!sufix.fail && !sufix.null) {
     parts = parts.concat(sufix.parts);
   }
-  if (config.debug.spawn) {
+  if (config.debug.spawn && !countOnly) {
```
TooAngel: Before introducing a new parameter, I would rephrase the log message, like replacing `Spawning` with `getPartConfig`.

It is just a debug statement.

samogot: In that case, there will be wall of log messages of this type and that become meaningless, 
I'd rather refactor this function and move log message about spawning in place of real spawning

samogot:
Yeah, `spawnReplacement` is already good enough for me. It only take into account time of last spawning, but not compute time of spawning next creep, but as constructing new extensions is'nt often in later stages of game - real spawning time can be neglected

PR #253, opened 2017-06-02
Improve logic on sourcer overflow. 

samogot:
Improve logic on sourcer overflow. Instead of waiting for dropped energy on each tick, until creep get full, withdraw than energy from sourcer/container, to get full in two ticks and free up capacity for sourcer to avoid overflowing

### DIFF
--- a/src/prototype_creep_startup_tasks.js
+++ b/src/prototype_creep_startup_tasks.js
@@ -336,6 +336,39 @@ Creep.prototype.repairStructure = function() {
   return false;
 };
 
+/**
+ *
+ * @param {Resource} target Resource object to pick up
+ * @return {number} total received resources amount
+ */
+Creep.prototype.pickupOrWithdrawFromSourcer = function(target) {
+  const creepFreeSpace = this.carryCapacity - _.sum(this.carry);
+  let pickedUp = 0;
+  // this.log('pickupOrWithdrawFromSourcer free '+creepFreeSpace+' '+target+' '+target.amount)
+  if (target.amount < creepFreeSpace) {
+    let container = target.pos.lookFor(LOOK_STRUCTURES).find(function(structure) {
+      return structure.structureType === STRUCTURE_CONTAINER && structure.store[target.resourceType] > 0 && _.sum(structure.store) === structure.storeCapacity;
+    });
+    if (container) {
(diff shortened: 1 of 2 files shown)


democratic-collaboration:

Please review the PR to help.

src/prototype_creep_startup_tasks.js
```diff
@@ -336,6 +336,39 @@ Creep.prototype.repairStructure = function() {
...
+        return creep.memory && creep.memory.role === 'sourcer' && creep.carry[target.resourceType] > 0 && _.sum(creep.carry) === creep.carryCapacity;
+      });
+      if (sourcer) {
+        const toWithdraw = Math.min(creepFreeSpace - target.amount, sourcer.carry[target.resourceType]);
+        sourcer.transfer(this, target.resourceType, toWithdraw);
+        pickedUp += toWithdraw;
+      }
+    }
+  }
+  this.pickup(target);
```
TooAngel: Not sure how the ordering in action will be. But I would do the `pickup()` before the `withdraw()`. Energy on the ground decays, while in sourcer or container will stay. Will not change a lot, but a bit better.

democratic-collaboration:
A new review, yeah.

    Votes: 181/358
    Coefficient: 0.505586592179
    Merging in 7 days 6 hours
    Age 0 days 3 hours

src/prototype_creep_startup_tasks.js
samogot: Acording to http://docs.screeps.com/simultaneous-actions.html `pickup` isn't involved into any of pipelins, so it should be done in one tick with `withdraw`.
Nevertheless, if we do `pickup` in first tick and `withdraw` in second, it woun't work, as on the second tick we will already have new 10 energy for picking up. 1 energy unit isn't worth much code complication

TooAngel: I thought more about within one tick: first pickup, then withdraw somehow missed the `toWithdraw` variable. Fine, well done. Ignore what I'm saying ;-)

Btw. if you like come join the #angels_public channel on slack

democratic-collaboration:
A new review, yeah.

    Votes: 181/358
    Coefficient: 0.505586592179
    Merging in 7 days 6 hours
    Age 0 days 3 hours

PR #269, opened 2017-06-23
Try to add more clever prioritizing in getEnergy between storage and …

samogot:
…dropped energy

Get dropped energy first only if it's amount is bigger than amount of energy in storage, bigger than carry capacity of creep or if dropped energy is twice nearer to creep than storage

### DIFF
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -470,11 +470,18 @@ Creep.prototype.getEnergy = function() {
     return false;
   }
 
-  if (this.getDroppedEnergy()) {
-    return true;
+  if (this.room.storage) {
+    if (this.getDroppedEnergy(res => res.amount > this.carryCapacity ||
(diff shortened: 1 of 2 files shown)


src/prototype_creep_resources.js
```diff
@@ -470,14 +470,21 @@ Creep.prototype.getEnergy = function() {
     return false;
   }
 
-  if (this.getDroppedEnergy()) {
-    return true;
+  if (this.room.storage) {
+    if (this.getDroppedEnergy(res => res.amount > this.carryCapacity ||
+        res.amount > this.room.storage.store.energy || this.pos.getRangeTo(res) * 2 < this.pos.getRangeTo(this.room.storage))) {
+      return true;
+    }
```
TooAngel: Hm, I don't like this so much.

I think we should get the droppedEnergy first, if it makes sense. Especially because droppedEnergy decays, energy in the storage not. If not get it from the storage.
Instead of working with conditions here, I would prefer if we could tweak the formulas in 
`src/prototype_creep_startup_tasks.js` line 384.

So the idea is to have some kind of calculation if it makes sense to move to that droppedEnergy.

PR #287, opened 2017-07-08
Fix slack channel name and provide credentials for the graphs on scre…

TooAngel:
…epspl.us

### DIFF
--- a/config_local.js.example
+++ b/config_local.js.example
@@ -1,6 +1,6 @@
 // Local overrides to config, modify this instead of global config.
-//--- Uncomment screepsplusToken for share datas on TooAngels datasboard on screepspl.us
-//--- You can acces datas contacting tooangel on #the_angels slack channel or use your token.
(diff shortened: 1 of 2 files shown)


samogot:
Is it a good point to share credentials in public? This way anyone, not only Angels contributor or user can read stats, so our players would be vulnerable to other

TooAngel:
Oh, didn't recognize that you started using it. This was a long discussion in the past, that we wanted to start sharing them internally first and provide them public later. In the sense of we have an open source code base, we can also make our graphs and maybe logs open source. - And if so, we can still win :-)

In the end it is open for discussion. If you prefer I can close this and change the password.

TooAngel:
I often had the situation where I would love to see code from other players or get a feeling of their resources in graphs or logs. I'm pretty much wanting to go in the direction of completely open and a contributor based development.
Let's say taking me a bit out of the equation for the future development of the TooAngel account.

samogot:
ok, that makes sence

PR #290, opened 2017-07-08
[WIP] Introduce enforcement of documentation changes

TooAngel:
I think the project lacks a lot of documentation, I guess most people ran into that problem.

To enforce improvement of the documentation the CI checks will also check if a markdown file has changed and will fail if not.

It surely will help me to work constantly on the documentation. Being honest each contribution should have a documentation improvement in there anyway, at least in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+ - introduce documenation changes in the tests
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -23,6 +23,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-uglify');
   grunt.loadNpmTasks('grunt-sync');
   grunt.loadNpmTasks('grunt-contrib-clean');
+  grunt.loadNpmTasks('grunt-exec');
 
   grunt.initConfig({
     screeps: {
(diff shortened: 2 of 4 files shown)


democratic-collaboration:

Please review the PR to help.

democratic-collaboration:
Adding WIP label, the title is prefixed with [WIP]

samogot:
I'm not sure this is best decision. A lot of small changes do not need a lot of documentation and are self-explanatory by commit message. Enforcing contributors to copy-paste commit messages into changelog will decrease motivation to contribute. Also appending lines to the end of changelog from different PRs will lead to merge conflicts.

I agree that documentation is poor, but what we neet is not to enforce useless changlogs. We rather need detailed description of AI logic in complex and/or in-code documentation of methods using JSDoc

Only if we will have it - enforcing to modify *.md files would make some sence. I mean it will be usefull to maintain description in actual state in case of logic changes, but still not all PRs are changes in logic

TooAngel:
It is still a thought process, thank you for the feedback.

The TooAngel AI should be more often release to npm, so that it will be available as bot on the private screeps server. So recently I introduce the automatic publish to npm as soon as the version changes.
So next step would be to create a changelog. In the early days I had to copy all the commit messages to the manual changelog. This could be done by the contributors, too?

Getting merge conflicts in the PRs is a valid and good point.

The point about JSDoc is good, too. I thought mostly about the github / markdown documentation. What about either documentation or JSDoc changes?

samogot:
I mean that if description in JSDoc would be well detailed and include enough logic description, it may replace separate markdown documentation. I'm not sure about genegating html docs from JSDoc, because personally I will not use it) My IDE handles JSDoc pretty good, and when it isn't enough for me it's faster to jump to function definition and read documentation and/or source there, instead of opening some external docs.

TooAngel:
Ok, I take you point.

I think there are different layers of documentation:
The markdown high level, best extended with images, documentation. To describe the overall concepts.
And the jsDoc for describing the functions and logic.

I think both needs to be improved a lot.

So my main point here is, how can I make sure that this will improve over time and will stay at that level?
In the end I think it needs to be enforced by some kind of checks. Any suggestions, anyone?

And still the topic about how to have a changelog is open.

mschultheiss83:
minus for changelog

but

plus for jsDoc
plus for md-file for each BIG topic e.g. Remote Rooms / Mineral Handling / Boosting / Keepers and so on
plus for better config documentation

Issue #190, opened 2017-02-05
Carry Issue 4


totalschaden:
http://i.imgur.com/M06kEVo.png

Problem:
Carry 829 and 833 have the Source that is nearby as TargetId. 
They will switch Positions back and forth as long as none of them is considerd full, blocking all Carrys that have to go on on the Path to another Source.

Possible Solution: 
Dont let Carrys wait for Energy on the main path. We could let them wait at the Link Position instead ?! (min 2 range from Path?)

invisioblack:
couldn't we apply the stuck logic to the carries and have them try to move around by MoveTo(ing) to a point further up the path if isStuck() is true?

Possibly something like this?


[StuckExample.txt](https://github.com/TooAngel/screeps/files/753045/StuckExample.txt)

TooAngel:
I think the issue is, that one carry is waiting for enough energy from the sourcer. Maybe the carry should move a step back on the path, each time it gets energy from the sourcer/container.

TooAngel:
Also some improvement in the last month here.

Issue #126, opened 2017-01-10
Carry Issue 2


totalschaden:
theres a problem when some carrys have a long way, and on that way there are like 3 sources but they need to go to the 4th, then all time time they get handed engery from carrys that come from source 1,2 or 3 so they never reach source 4

TooAngel:
Some small fixes in my rooms it evens out

PR #302, opened 2017-07-18
Improve sourcer & carry handling

TooAngel:
- Make sizes dependant on the capacityAvailable
 - Calculate carry spawn based on the size and throughput of the sourcer over time

Some fixes
 - Improve carry handling, stuck on RCL 4
 - Fix external room roads
 - Fix reserver

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -69,7 +69,7 @@ global.config = {
     scoutMinControllerLevel: 4,
     ttlPerRoomForScout: 500,
     numberOfNextroomers: 10,
-    nextroomerInterval: _.ceil(1500 / 10),
+    nextroomerInterval: 500,
     maxRooms: 20,
     revive: true,
     maxDistance: 17,
@@ -118,18 +118,19 @@ global.config = {
     distance: 3,
   },
 
-  sourcer: {
-    spawnCarryLevelMultiplier: 300,
-    spawnCarryWaitTime: 400,
-  },
-
   carry: {
-    size: 200,
+    sizes: {
+      0: [3, 3], // RCL 1
+      550: [4, 4], // RCL 2
+      800: [6, 6], // RCL 3
+      1300: [6, 11], // RCL 4
+      1800: [8, 15], // RCL 5
+      2300: [11, 21], // RCL 6
+    },
     // Percentage should increase from base to target room. Decrease may cause stack on border
     carryPercentageBase: 0.1,
     carryPercentageHighway: 0.2,
-    carryPercentageExtern: 0.5,
-    maxPerTargetPerRoom: 2,
+    carryPercentageExtern: 0.5
   },
 
   creep: {
@@ -196,16 +197,17 @@ global.config = {
       harvester: 1,
       sourcer: 2,
       storagefiller: 3,
-      defendranged: 3
+      defendranged: 4,
+      carry: 5
     },
     otherRoom: {
       harvester: 11,
       defender: 12,
       defendranged: 13,
       nextroomer: 15,
-      reserver: 16,
       carry: 17,
-      sourcer: 18
+      sourcer: 18,
+      reserver: 19
     }
   }
 };
--- a/src/prototype_creep.js
+++ b/src/prototype_creep.js
@@ -167,9 +167,12 @@ Creep.prototype.stayInRoom = function() {
 };
 
 Creep.prototype.buildRoad = function() {
-  if (this.room.controller && this.room.controller.my && this.pos.lookFor(LOOK_TERRAIN)[0] !== 'swamp' &&
-    (this.room.controller.level < 4 || this.room.memory.misplacedSpawn)) {
-    return false;
+  if (this.room.controller && this.room.controller.my) {
+
+    if (this.pos.lookFor(LOOK_TERRAIN)[0] !== 'swamp' &&
+      (this.room.controller.level < 3 || this.room.memory.misplacedSpawn)) {
+      return false;
+    }
   }
 
   // TODO as creep variable
--- a/src/prototype_creep_harvest.js
+++ b/src/prototype_creep_harvest.js
@@ -43,8 +43,18 @@ Creep.prototype.handleSourcer = function() {
 };
 
 Creep.prototype.spawnCarry = function() {
-  var energyThreshold = Game.rooms[this.memory.base].controller.level * config.sourcer.spawnCarryLevelMultiplier;
-  var waitTime = config.sourcer.spawnCarryWaitTime;
+  if (this.memory.wait > 0) {
+    this.memory.wait -= 1;
+    return false;
+  }
+
+  const foundKey = Object.keys(config.carry.sizes).reverse()
+    .find(key => (key <= Game.rooms[this.memory.base].energyCapacityAvailable));
+  let carryCapacity = config.carry.sizes[foundKey][1] * CARRY_CAPACITY;
+
+  const workParts = this.body.filter(part => part.type === WORK).length;
+
+  let waitTime = carryCapacity / (HARVEST_POWER * workParts);
 
   var spawn = {
     role: 'carry',
@@ -54,36 +64,20 @@ Creep.prototype.spawnCarry = function() {
     }
   };
 
-  // Spawn carry
-  var energies = this.pos.lookFor(LOOK_ENERGY);
-  if (energies.length === 0) {
-    const containers = this.pos.findInRangePropertyFilter(FIND_STRUCTURES, 0, 'structureType', [STRUCTURE_CONTAINER], false, {
-      // TODO hardcoded for now, half of the container? Good idea?
-      filter: container => container.store.energy >= 1000
-    });
-    if (containers.length === 0) {
-      return false;
-    }
+  let resourceAtPosition = 0;
+  var resources = this.pos.lookFor(LOOK_RESOURCES);
+  for (let resource of resources) {
+    resourceAtPosition += resource.amount;
   }
 
-  if (energies.length > 0 && energies[0].amount < 50) {
-    return false;
+  let containers = this.pos.findInRangeStructures(FIND_STRUCTURES, 0, STRUCTURE_CONTAINER);
+
+  for (let container of containers) {
+    resourceAtPosition += _.sum(container.store);
   }
 
-  if (this.inBase()) {
-    if (energies.length > 0 && energies[0].amount < energyThreshold) {
-      return false;
-    }
+  if (resourceAtPosition > carryCapacity) {
+    Game.rooms[this.memory.base].checkRoleToSpawn('carry', 0, this.memory.routing.targetId, this.memory.routing.targetRoom);
   }
-
-  if (!existInArray(Game.rooms[this.memory.base].memory.queue, spawn)) {
-    if (typeof(this.memory.wait) === 'undefined') {
-      this.memory.wait = 0;
-    }
-    if (this.memory.wait <= 0) {
-      Game.rooms[this.memory.base].checkRoleToSpawn('carry', config.carry.maxPerTargetPerRoom, this.memory.routing.targetId, this.memory.routing.targetRoom);
-      this.memory.wait = waitTime;
-    }
-  }
-  this.memory.wait -= 1;
+  this.memory.wait = waitTime;
 };
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -304,7 +304,12 @@ Room.prototype.executeRoom = function() {
       if (this.controller.level === 2 || this.controller.level === 3) {
         amount = 5;
       }
+    } else {
+      if (this.storage.store.energy < config.creep.energyFromStorageThreshold && this.controller.level < 5) {
+        amount = 3;
+      }
     }
+
     this.checkRoleToSpawn('harvester', amount, 'harvester');
   }
 
--- a/src/role_carry.js
+++ b/src/role_carry.js
@@ -16,16 +16,14 @@ roles.carry.flee = true;
 roles.carry.boostActions = ['capacity'];
 
 roles.carry.settings = {
-  param: ['controller.level'],
+  param: ['energyCapacityAvailable'],
   prefixString: {
-    1: 'MC',
-    3: 'WMC',
+    300: '',
+    550: 'W'
   },
   layoutString: 'MC',
-  amount: {
-    1: [1, 1],
-    3: [1, 2]
-  }
+  amount: config.carry.sizes,
+  maxLayoutAmount: 1
 };
 
 roles.carry.checkHelperEmptyStorage = function(creep) {
--- a/src/role_sourcer.js
+++ b/src/role_sourcer.js
@@ -14,25 +14,24 @@
 roles.sourcer = {};
 
 roles.sourcer.settings = {
-  param: ['controller.level'],
+  param: ['energyCapacityAvailable'],
   prefixString: {
-    1: 'MW',
-    3: 'MWC'
+    300: 'MW',
+    600: 'MWC'
   },
   layoutString: {
-    2: 'W',
-    3: 'MW',
-    5: 'MW'
+    300: 'W',
+    650: 'MW'
   },
   amount: {
-    2: [1],
-    3: [1, 1],
-    5: [2, 4]
+    300: [1],
+    650: [1, 4],
+    700: [2, 4]
   },
   maxLayoutAmount: {
-    1: 4,
-    5: 1,
-  },
+    300: 4,
+    650: 1
+  }
 };
 
 roles.sourcer.buildRoad = true;
--- a/.jshintrc
+++ b/.jshintrc
@@ -134,7 +134,10 @@
     "RawVisual": false,
     "LOOK_CREEPS": false,
     "STRUCTURE_KEEPER_LAIR": false,
-    "OBSTACLE_OBJECT_TYPES": false
+    "OBSTACLE_OBJECT_TYPES": false,
+    "CARRY_CAPACITY": false,
+    "HARVEST_POWER": false,
+    "LOOK_RESOURCES": false
   },
   "node": true
 }
--- a/src/prototype_roomPosition.js
+++ b/src/prototype_roomPosition.js
@@ -32,7 +32,7 @@ RoomPosition.prototype.clearPosition = function(target) {
 };
 
 RoomPosition.prototype.getClosestSource = function(filter) {
-  let source = this.findClosestByPath(FIND_SOURCES_ACTIVE, {filter});
+  let source = this.findClosestByPath(FIND_SOURCES_ACTIVE, { filter });
   if (source === null) {
     source = this.findClosestByRange(FIND_SOURCES_ACTIVE);
   }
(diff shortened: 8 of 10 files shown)


src/prototype_creep.js
```diff
@@ -167,9 +167,16 @@ Creep.prototype.stayInRoom = function() {
...
-    (this.room.controller.level < 4 || this.room.memory.misplacedSpawn)) {
-    return false;
+  if (this.room.controller && this.room.controller.my) {
+
+    if (this.pos.lookFor(LOOK_TERRAIN)[0] !== 'swamp' &&
+      (this.room.controller.level < 3 || this.room.memory.misplacedSpawn)) {
+      return false;
+    }
+    // TODO should be extracted to a method, stolen from role_harvester
+    if (!this.room.storage || this.room.storage.store.energy + 1000 < config.creep.energyFromStorageThreshold) {
```
samogot: `this.room.controller.level < 3` always means that `!this.room.storage`, so we again wouldn't have roads on swamps in low RCL

Also I'm not sure this approach is good at all. I nearly always have low energy in storage, as upgrader spent it even faster then harvester fills extentions. We do not need a lot of energy for repairing roads, but if it'll decay, we'll fill storage even slower. So repair is mast have.
Do not build roads with low energy? May be... But I afraid that in low RCL rooms without other RCL 8 energy donor, compleat building roads will take much more time that it should.

src/prototype_creep_harvest.js
```diff
@@ -43,8 +43,28 @@ Creep.prototype.handleSourcer = function() {
...
 Creep.prototype.spawnCarry = function() {
-  var energyThreshold = Game.rooms[this.memory.base].controller.level * config.sourcer.spawnCarryLevelMultiplier;
-  var waitTime = config.sourcer.spawnCarryWaitTime;
+  if (this.memory.wait > 0) {
+    this.memory.wait -= 1;
+    return false;
+  }
+
+  var foundKey;
+  for (let key of Object.keys(config.carry.sizes).sort(function(a, b) { return parseInt(a, 10) - parseInt(b, 10); })) {
```
samogot: `-` automatically converts string to int. like `("3" - "2") === 1`

src/prototype_creep_harvest.js
```diff
@@ -43,8 +43,28 @@ Creep.prototype.handleSourcer = function() {
...
+    foundKey = key;
+  }
+  let carryCapacity = config.carry.sizes[foundKey][0] * CARRY_CAPACITY;
+
+  var work_parts = 0;
+  for (var part_i in this.body) {
+    if (this.body[part_i].type === 'work') {
+      work_parts++;
+    }
+  }
```
samogot: const work_parts = this.body.filter(part => part.type === WORK).length;

src/prototype_creep_harvest.js
samogot: moreover, do we really need sorting here each time?

src/prototype_creep_harvest.js
```diff
@@ -54,36 +74,33 @@ Creep.prototype.spawnCarry = function() {
...
-    const containers = this.pos.findInRangePropertyFilter(FIND_STRUCTURES, 0, 'structureType', [STRUCTURE_CONTAINER], false, {
-      // TODO hardcoded for now, half of the container? Good idea?
-      filter: container => container.store.energy >= 1000
-    });
-    if (containers.length === 0) {
-      return false;
-    }
+  let energyAtPosition = 0;
+  var energies = this.pos.lookFor(LOOK_RESOURCES);
+  for (let energy of energies) {
```
samogot: check for resourceType? or rename variable

src/prototype_creep_harvest.js
```diff
@@ -54,36 +74,33 @@ Creep.prototype.spawnCarry = function() {
...
+    energyAtPosition += energy.amount;
   }
 
-  if (energies.length > 0 && energies[0].amount < 50) {
-    return false;
+  if (energyAtPosition > carryCapacity) {
+    Game.rooms[this.memory.base].checkRoleToSpawn('carry', config.carry.maxPerTargetPerRoom, this.memory.routing.targetId, this.memory.routing.targetRoom);
```
samogot: do we need `config.carry.maxPerTargetPerRoom` limit at all in this approach?

src/prototype_creep_harvest.js
```diff
@@ -54,36 +74,33 @@ Creep.prototype.spawnCarry = function() {
...
+    this.memory.wait = waitTime;
+    return true;
   }
 
-  if (this.inBase()) {
-    if (energies.length > 0 && energies[0].amount < energyThreshold) {
-      return false;
+  let containers = this.pos.findInRange(FIND_STRUCTURES, 0, {
```
samogot: findInRangePropertyFilter?

src/prototype_creep_harvest.js
```diff
@@ -54,36 +74,33 @@ Creep.prototype.spawnCarry = function() {
...
-  if (!existInArray(Game.rooms[this.memory.base].memory.queue, spawn)) {
-    if (typeof(this.memory.wait) === 'undefined') {
-      this.memory.wait = 0;
-    }
-    if (this.memory.wait <= 0) {
-      Game.rooms[this.memory.base].checkRoleToSpawn('carry', config.carry.maxPerTargetPerRoom, this.memory.routing.targetId, this.memory.routing.targetRoom);
-      this.memory.wait = waitTime;
-    }
+  if (energyAtPosition > carryCapacity) {
+    Game.rooms[this.memory.base].checkRoleToSpawn('carry', config.carry.maxPerTargetPerRoom, this.memory.routing.targetId, this.memory.routing.targetRoom);
```
samogot: why duplicating for containers and for resources separately?

src/prototype_room_my.js
```diff
@@ -343,14 +347,12 @@ Room.prototype.executeRoom = function() {
...
-        role: 'defendranged'
-      };
+      let role = 'defendranged';
       if (this.memory.attackTimer > 300) {
-        defender.role = 'defendmelee';
+        role = 'defendmelee';
       }
-      if (this.exectueEveryTicks(250) && !this.inQueue(defender)) {
-        this.memory.queue.push(defender);
+      if (this.exectueEveryTicks(250)) {
```
samogot: what connection it makes for sourcers/carry logic? but ok))

src/role_carry.js
```diff
@@ -16,15 +16,14 @@ roles.carry.flee = true;
...
-    1: 'MC',
-    3: 'WMC',
+    800: 'W'
   },
   layoutString: 'MC',
-  amount: {
-    1: [1, 1],
-    3: [1, 2]
+  amount: config.carry.sizes,
+  maxLayoutAmount: {
```
samogot: why not `maxLayoutAmount: 1`?

src/role_sourcer.js
```diff
@@ -14,25 +14,24 @@
...
-    1: 'MW',
-    3: 'MWC'
+    150: 'MW',
+    600: 'MWC'
   },
   layoutString: {
-    2: 'W',
-    3: 'MW',
-    5: 'MW'
+    0: 'W',
```
samogot: 0 is overhead for energyCapacityAvailable. 300 is minimum value

democratic-collaboration:
A new review, yeah.

    Votes: 168/407
    Coefficient: 0.412776412776
    Merging in 5 days 19 hours
    Age 0 days 1 hours

src/prototype_creep_harvest.js
samogot: maybe even something like this will be better https://github.com/samogot/screeps/blob/master/dist/extendPrototypes.js#L34-L53

src/config.js
```diff
@@ -118,13 +118,15 @@ global.config = {
...
-
   carry: {
-    size: 200,
+    sizes: {
+      0: [3, 3], // RCL 1
+      550: [5, 5], // RCL 2
+      800: [6, 6], // RCL 3
+      1300: [10, 10], // RCL 4
+      1800: [14, 14], // RCL 5
+      2300: [20, 20], // RCL 6
```
TooAngel: Good point, thx

src/prototype_creep.js
TooAngel: K, removed again. I had harvester running around and wanted to prevent that. I think the better approach is to fix the harvester pathing.

Overall I think roads bring more benefits than disadvantages. I'm improving the `upgrader` settings a bit in my workspace branch. I think we can control the storage energy level, by adapting the upgrader.

src/prototype_creep_harvest.js
TooAngel: thx

src/prototype_creep_harvest.js
TooAngel: thx

src/prototype_creep.js
samogot: I do the same in my workspace too. I mean I also restrict upgrader appetite, but have no clue is we se the same approach)

src/prototype_creep_harvest.js
TooAngel: no

src/prototype_creep_harvest.js
TooAngel: If there is enough energy, we don't have to find the container, save CPU

src/role_carry.js
TooAngel: thx

src/prototype_room_my.js
TooAngel: true, ran into the problem, while working on the carry things, will extract

src/role_sourcer.js
TooAngel: Not sure what you mean. from 0 to 649 the `layoutString` should be `'W'`, well I could also put `150`, but `0` makes clear from the beginning.

src/prototype_creep_harvest.js
samogot: Even as I have CPU limit problems now, I still prever neat code more)) Does it really makes difference once per waitTime?

src/role_sourcer.js
samogot: 300 is capacity of spawn. we allways have spawn, so 300 is minimum value of energyCapacityAvailable

src/prototype_creep_harvest.js
TooAngel: regarding converting I ran exactly into the problem that it was wrong. Had the same issue with different browser today.

Possible that the behavior changed recently.

I'm not an nodejs expert. Is the sorting executed for each element again or only once. I can move it to a variable before?!

src/prototype_creep_harvest.js
TooAngel: thx

src/role_sourcer.js
TooAngel: Ok, I think it doesn't matter ... somehow.

True 300 energy is the default minimum, but it wouldn't matter to use 0, correct?
Usually, I see private server where some constants are increased, but you could also decrease it, let's say spawn capacity 200, which would break it ....

src/prototype_creep_harvest.js
samogot: sorting executed only once if you mean for loop, but it is executed on each function run (every `waitTime` ticks). Why we should do it? Can't we enforce that settings should already be sorted?

TooAngel: Well it is an object not an array AFAIK they are not sorted, so here it gets a bit more tricky. I think we can leave it like this right now. When checking the performance next time either it pops up again or it is not so bad.

src/role_sourcer.js
samogot: yeah, didn't think about it

TooAngel:
I squash the commit, all comments were addressed, at least.

src/prototype_creep_harvest.js
samogot: properties in JS objects are sorted according to "original insertion order". In our case of object literal, order would be the same as in our literal

samogot: the only posibility why we may need sort here - if someone redefined this setting in his config.local with random property order. Or if in future we'll forget about order is matter and will somehow modify setting object's propperies set after original initialisation

src/prototype_creep_harvest.js
```diff
@@ -49,8 +49,7 @@ Creep.prototype.spawnCarry = function() {
   }
 
   var foundKey;
-  let sizes = Object.keys(config.carry.sizes).sort(function(a, b) { return parseInt(a, 10) - parseInt(b, 10); });
-  for (let key of sizes) {
+  for (let key in config.carry.sizes) {
```
TooAngel: @samogot So mean just like this?

samogot: yep. `for (let key of Object.keys(config.carry.sizes))` is good aswell. `Object.keys` will enumerate only own properties, and so it do not trigger warning in codeclimate (in this it's suppressed, but logicaly warning should still be there)

src/prototype_creep_harvest.js
```diff
@@ -43,8 +43,23 @@ Creep.prototype.handleSourcer = function() {
...
+    return false;
+  }
+
+  var foundKey;
+  for (let key in config.carry.sizes) {
+    if (Game.rooms[this.memory.base].energyCapacityAvailable < key) {
+      break;
+    }
+    foundKey = key;
+  }
```
samogot: not better in performance, but IMHO more clear one-liner
```
const foundKey = Object.keys(config.carry.sizes)
  .filter(key=> key <= Game.rooms[this.memory.base].energyCapacityAvailable).pop()
```

democratic-collaboration:
A new review, yeah.

    Votes: 228/407
    Coefficient: 0.560196560197
    Merging in 6 days 13 hours
    Age 0 days 0 hours

src/prototype_creep_harvest.js
TooAngel: Ok, cool, thanks, will change.

Why is it suppressed?

samogot: Because we have if in for-in loop. codeclimate suposed that if we have if, then we validate key

The error might be if in some moment in future, someone decide to modify `Object.prototype` for example to add some usefull method. In that case, this "usefull" method may appear in all for-in loops as key, so if someone use for-in loop, it's better to check hasOwnProperty, just in case)

democratic-collaboration:
A new review, yeah.

    Votes: 228/407
    Coefficient: 0.560196560197
    Merging in 10 days 23 hours
    Age 0 days 0 hours

src/prototype_creep_harvest.js
```diff
@@ -43,8 +43,18 @@ Creep.prototype.handleSourcer = function() {
...
 
 Creep.prototype.spawnCarry = function() {
-  var energyThreshold = Game.rooms[this.memory.base].controller.level * config.sourcer.spawnCarryLevelMultiplier;
-  var waitTime = config.sourcer.spawnCarryWaitTime;
+  if (this.memory.wait > 0) {
+    this.memory.wait -= 1;
+    return false;
+  }
+
+  const foundKey = Object.keys(config.carry.sizes).reverse()
```
Nevrdid: or with only only looping throught object : https://github.com/TooAngel/screeps/blob/carrysourcer/src/prototype_room_creepbuilder.js#L211

src/prototype_creep_harvest.js
```diff
@@ -43,8 +43,18 @@ Creep.prototype.handleSourcer = function() {
...
+    .find(key => (key <= Game.rooms[this.memory.base].energyCapacityAvailable));
+  let carryCapacity = config.carry.sizes[foundKey][1] * CARRY_CAPACITY;
+
+  const workParts = this.body.filter(part => part.type === WORK).length;
```
Nevrdid: we can ally looping until we encounter a different part type as they are sorted https://github.com/TooAngel/screeps/blob/carrysourcer/src/prototype_room_creepbuilder.js#L307
depending if we want to definitly sort them or still allow to unsort. but we still can place them at start or end.

democratic-collaboration:
A new review, yeah.

    Votes: 283/407
    Coefficient: 0.695331695332
    Merging in 2 days 22 hours
    Age 0 days 2 hours

src/prototype_room_my.js
```diff
@@ -303,6 +303,10 @@ Room.prototype.executeRoom = function() {
       // TODO maybe better spawn harvester when a carry recognize that the dropped energy > threshold
       if (this.controller.level === 2 || this.controller.level === 3) {
         amount = 5;
+      } else {
+        if (this.storage.store.energy < config.creep.energyFromStorageThreshold && this.controller.level < 5) {
+          amount = 3;
+        }
```
samogot: missplaced. in this if-branch `this.storage` is undefined

TooAngel: true thx, some messed up the squash rebase

src/prototype_creep_harvest.js
TooAngel: We had that before, I think this is more the javascript way

src/prototype_creep_harvest.js
TooAngel: I wouldn't rely on sorted body parts. If we change that at some point in time, it is really hard to find such issues.

src/prototype_creep_harvest.js
```diff
@@ -54,36 +64,20 @@ Creep.prototype.spawnCarry = function() {
...
-  if (!existInArray(Game.rooms[this.memory.base].memory.queue, spawn)) {
-    if (typeof(this.memory.wait) === 'undefined') {
-      this.memory.wait = 0;
-    }
-    if (this.memory.wait <= 0) {
-      Game.rooms[this.memory.base].checkRoleToSpawn('carry', config.carry.maxPerTargetPerRoom, this.memory.routing.targetId, this.memory.routing.targetRoom);
-      this.memory.wait = waitTime;
-    }
+  if (resourceAtPosition > carryCapacity) {
+    Game.rooms[this.memory.base].checkRoleToSpawn('carry', config.carry.maxPerTargetPerRoom, this.memory.routing.targetId, this.memory.routing.targetRoom);
```
samogot: 0 instead of `config.carry.maxPerTargetPerRoom`
also squash issue?)

samogot:
I found a problem. This logic is not very applicable for big resource piles. In that case, sourcer will spawn a new carry every 20 ticks, until resource pile is gone, but we do not need so much creeps for that. Some of latest spawned creeps wouldn't even get to the sourcer until resource pile is gone. And after that total carry capacity per sourcer would be much more then sourcer power. I guess they will stand up in the line and wait near the sourcer. This is huge waste of energy. We should either find a way to stop spawning them in right moment, or at least recycle them after pile is gone.

src/prototype_creep_harvest.js
TooAngel: k, thx

TooAngel:
> I found a problem. This logic is not very applicable for big resource piles. In that case, sourcer will spawn a new carry every 20 ticks, until resource pile is gone, but we do not need so much creeps for that. Some of latest spawned creeps wouldn't even get to the sourcer until resource pile is gone. And after that total carry capacity per sourcer would be much more then sourcer power. I guess they will stand up in the line and wait near the sourcer. This is huge waste of energy. We should either find a way to stop spawning them in right moment, or at least recycle them after pile is gone.

I'm not sure if it is an issue. First of all the pile of energy should not come up, why does that happen? When switching to this branch yes, can happen, but usually not.

Maybe introducing a trend for the resources and adapt the wait time depending on that.

I think this PR improves the sourcer / carry handling, further improvement can / should go into a new PR.

samogot:
For example resource pile might appear because of #126 or #190. Real problem of pile appearing was because of carry limit, but nevertheless I faced both of them in new room after respawn.

PR #307, opened 2017-07-20
if there link near to sourcer, try to pickup dropped energy…

samogot:
…and transfer it to link

### DIFF
--- a/src/prototype_creep_harvest.js
+++ b/src/prototype_creep_harvest.js
@@ -37,8 +37,14 @@ Creep.prototype.handleSourcer = function() {
       }
     }
 
-    let link = Game.getObjectById(this.memory.link);
-    this.transfer(link, RESOURCE_ENERGY);
+    const link = Game.getObjectById(this.memory.link);
+    if (link) {
+      this.transfer(link, RESOURCE_ENERGY);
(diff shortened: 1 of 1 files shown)


src/prototype_creep_harvest.js
```diff
@@ -37,8 +37,14 @@ Creep.prototype.handleSourcer = function() {
...
     }
 
-    let link = Game.getObjectById(this.memory.link);
-    this.transfer(link, RESOURCE_ENERGY);
+    const link = Game.getObjectById(this.memory.link);
+    if (link) {
+      this.transfer(link, RESOURCE_ENERGY);
+      const resources = this.pos.findInRangePropertyFilter(FIND_DROPPED_RESOURCES, 1, 'resourceType', [RESOURCE_ENERGY]);
+      if (resources.length > 0) {
+        this.pickup(resources);
```
TooAngel: Not sure about the cpu, but we can try.

Can you reduce the amount to pickup to `max(resources.amount, capacity - energyFromSourceInNextTick)`, so that we can still continue to harvest

samogot: we do not need carry space to harvest) energy would just dropped back)

TooAngel: true

democratic-collaboration:
A new review, yeah.

    Votes: 228/407
    Coefficient: 0.560196560197
    Merging in 4 days 8 hours
    Age 0 days 0 hours

PR #308, opened 2017-07-20
restrict upgrader energy consumption

samogot:


src/prototype_creep_resources.js
```diff
@@ -181,9 +181,12 @@ Creep.prototype.handleUpgrader = function() {
...
 
-  returnCode = this.withdraw(this.room.storage, RESOURCE_ENERGY);
-  if (returnCode === ERR_FULL || returnCode === OK) {
-    return true;
+  if ((this.room.energyCapacityAvailable - this.room.energyAvailable + CARRY_CAPACITY < this.room.storage.store.energy &&
+      this.room.memory.queue.length === 0) || this.room.controller.ticksToDowngrade < 3000) {
+    returnCode = this.withdraw(this.room.storage, RESOURCE_ENERGY);
+    if (returnCode === ERR_FULL || returnCode === OK) {
+      return true;
+    }
```
TooAngel: Hm, not yet convinced.

I wouldn't make so much connection to different energy states, especially the queue. The `upgrader` has not really something to do with the energyAvailable.
We could make it dependant on the `storage.store.energy > someValue` where someValue is in the config or double the harvester value. (Keep the ticksToDowngrade)
I was also more thinking about the `upgrader` size, e.g. check the storage when spawning and make sure it can only consume half of the energy in the storage. 1 work part consumes in 1500 tick 1500 energy. E.g. 10k energy in storage, so take half, 3 work parts.

Issue #89, opened 2017-01-05
Road construction sites below sourcers in baserooms never get finished. Remove/finish them?


Somotaw:

TooAngel:
Yes, true.

A role can set `buildRoad = true` which makes sure roads are build on the path. This is also true for the last position of a sources. After reaching the last pos and setting the constructionSite the sourcer sets `routing.reached = true` and will skip road building.

I think they don't need to build the road, because of no more moving. In my mind fixing this is a bit more tricky or ugly than just changing one line. Pain < Advantage = Didn't touch yet. Feel free, would be cool.

Somotaw:
adding to TODO list :)

mschultheiss83:
fixed in https://github.com/TooAngel/screeps/pull/315?

TooAngel:
> fixed in #315?

That would have been sooo cool :-) I still see them sometimes

mschultheiss83:
i did not found them in myRooms

mschultheiss83:
i think this is fixed see https://github.com/TooAngel/screeps/pull/498/files#diff-8676a6e81b3f4dc29af2900b07a60958R248
carries will build them up if there is a road construction site in range (<4)

TooAngel:
K, will close it

PR #315, opened 2017-07-27
Execute `buildRoad` only while on path

TooAngel:


stu43005:
If the room does not have a storage, harvester will not move on the path.

TooAngel:
> If the room does not have a storage, harvester will not move on the path.

true, `buildRoad` only drops in RCL > 3 and I have a workaround https://github.com/TooAngel/screeps/commit/de2840da6ae389c3b08761b6030bfb0c9f6c2135#diff-8676a6e81b3f4dc29af2900b07a60958R190

This is not the best solution I think `buildRoad` should only be executed when moving on the path.

TooAngel:
https://github.com/TooAngel/screeps/commit/cfb665e813d88e2e81658cf2ed4df3090982013d I think this will finally improve the situation. Would like to fix https://github.com/TooAngel/screeps/issues/89 at the same time, will send PR when I'm fine with it.

PR #317, opened 2017-07-27
Increase creep random number to reduce collisions

TooAngel:


stu43005:
Add a method to check the collision, and regenerate the name?

TooAngel:
> Add a method to check the collision, and regenerate the name?

When there is a collision the creep will not be spawned and tried again with a new number in the next tick. I could do the check first, increasing the scope is easier and less CPU intense. Any good reason to have short creep names?

democratic-collaboration:
A new review, yeah.

    Votes: 0/0
    Coefficient: 0
    Merging in 9 days 0 hours
    Age 0 days 23 hours

PR #322, opened 2017-07-30
Improve room layout-costmatix

samogot:
- Improve avoiding source keepers
- Avoid trivial swamps
- Avoid sources, so carries shouldn't pickup energy from different sourcer (related to #126 and #190)

### DIFF
--- a/src/prototype_room_costmatrix.js
+++ b/src/prototype_room_costmatrix.js
@@ -57,10 +57,21 @@ Room.prototype.setCostMatrixPath = function(costMatrix, path) {
     let pos = path[i];
     costMatrix.set(pos.x, pos.y, config.layout.pathAvoid);
   }
+  this.setCostMatrixAvoidSources(costMatrix);
 };
 
 Room.prototype.increaseCostMatrixValue = function(costMatrix, pos, value) {
   costMatrix.set(pos.x, pos.y, Math.max(costMatrix.get(pos.x, pos.y), value));
+};
+
+Room.prototype.setCostMatrixAvoidSources = function(costMatrix) {
+  const sources = this.find(FIND_SOURCES);
+  for (let source of sources) {
+    for (let pos of source.pos.getAllPositionsInRange(2)) {
+      this.increaseCostMatrixValue(costMatrix, pos, config.layout.sourceAvoid);
+    }
+  }
+  return sources;
 };
 
 Room.prototype.getCostMatrix = function() {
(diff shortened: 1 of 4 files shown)


src/prototype_room_costmatrix.js
```diff
@@ -57,41 +57,47 @@ Room.prototype.setCostMatrixPath = function(costMatrix, path) {
     let pos = path[i];
     costMatrix.set(pos.x, pos.y, config.layout.pathAvoid);
   }
+  this.setCostMatrixAvoidSources(costMatrix);
```
TooAngel: I think this is not necessary when you set it in `getCostMatrix`

src/prototype_room_costmatrix.js
```diff
@@ -57,41 +57,47 @@ Room.prototype.setCostMatrixPath = function(costMatrix, path) {
...
-        this.increaseCostMatrixValue(
-          costMatrix, {
-            x: lair.pos.x + dx,
-            y: lair.pos.y + dy
-          },
-          config.layout.skLairAvoid
-        );
+  if (lairs.length > 0) {
+    const minerals = this.find(FIND_MINERALS);
+    const sources = this.find(FIND_SOURCES);
```
TooAngel: This is similar to `setCostMatrixAvoidSources`, why do it twice.

also `skLairAvoidRadius` and `skLairAvoid` as config values doesn't fit so good.

I think I like this one more and we can remove `setCostMatrixAvoidSources`, what do you thinkg?

src/prototype_room_costmatrix.js
samogot: It is. CostMatrix path values overwrites values from `getCostMatrix`

src/prototype_room_costmatrix.js
samogot: Source Keeper may rather stay on lair or near source/mineral. `avoidSources` is used for other reasons.

src/prototype_room_costmatrix.js
TooAngel: Yes, true, but having a low avoid value on the path is fine. That the paths should avoid sources is handled when creating the path. When using `moveToMy` the creeps 'should' walk on the paths.
So if a creep wants to move to the source if should move on the path, if it moves somwhere else due to the surrounding high avoid values, it shouldn't move on that path anyway.
Makes sense?

src/prototype_room_costmatrix.js
TooAngel: Yes, true, well, maybe.
Both implementations avoid paths to be too close to special 'structures', for different reasons, with the same solution. The skLair with a greater radius (I hope), also solves the `setCostMatrixAvoidSources`. Or do I miss something?

samogot: It should. That was the original reason behind radius config option. But for some rooms bigger radius also blocks the only way through room, so pathFinder anyway takes shortest path which is near the keeper

TooAngel: That's also the same for sources?! If the room layout is the way that the path needs to go near the sources, it will.

The pathing solution is surely not the optimal but from my experience a robust, simple and not CPU intense one. There will be rooms where it doesn't work good, like the one you mention.

So we avoid positions close to the source because of carries, we avoid positions close to the special structures because of source keepers.
Does it harm if we avoid special structures in general. We could run into carry problems at minerals, too.
So avoiding special structures in general? If not set to 255 pathfinder will find the way if it needs to. The structures close to the mineral in owned rooms could have a bit more space if other paths diverge earlier. The code is easier.

So remove line 95, rename `sk` config variables, so that it happens in all rooms, would be my suggestion.

Do you get my point? Any good scenarios where the more fine grained avoidance is necessary?

samogot: Yeah, got it. What you are saying is acceptable solution. But I see no reasons why it is better than current one. Your proposal is more simple in logic and different config options, but it is the only advantage I can see. In general, the less restriction we apply onto path finding, the more optimal paths we get, so current solution is slightly preferable from energy point of view. Also your proposal has less complicared logic in "quality" aspect, but not in "quantity" as it will need more CPU computations - process costmatrix for positions around objects in cases we do not really need. So current solution is also slightly preferable from CPU point of view. And the last point - current solution is already implemented)) 
Any way either of three reasons have weak influence and can easily be ignored, but I see no reason to ignore them)

TooAngel: Ok, I think simple is a very strong argument, the game itself and the current code (structure) is not really simple, we should work on that. You also brought a similar argument in another PR, regarding simplicity to read.

Optimal path: It depends on the definition of the optimal path. How do I want to optimize it, distance, CPU usage to calculate or follow, energy consumption, ...., all of these are tricky and complex to compute. I think I'm mainly eyeing for robust (in terms of 'can be applied in most of the rooms') and simple. Hm, can't see a difference in both solutions, except more code = higher possibility of errors.

CPU computations: The costMatrix should be saved to cache or memory, so only computed rarely. I think `planer` and `repairer` (maybe `attacker` and `defender`) should be the only creeps which need to calculate new paths, which are in our rooms. 

Already implemented: That is a point yes and IMHO a very good argument.

Anyway, I'm thinking about removing line 95. Can't grep so argument so good, but I think it makes sense.

src/prototype_room_costmatrix.js
samogot: yeah, but if we don't make this change, path will not avoid sources. Only first one do, but after we build some path to sourcer and overwrite costmatrix, next paths won't avoid this source

TooAngel: source_avoid

Maybe let's discuss at an example, feel free to provide better discussion material, my 'color'-skills are limited.
The red dots are the `setCostMatrixAvoidSources` positions. The green dot is the only position where the source avoid costmatrix entry gets overwritten, which you want to reset. I think positions on paths should be by default useable for paths. Positions close the sources are not the most interesting ones for pathing because the surrounding positions should be avoided, except you want to move to the source.

Regarding the special structure avoidance topic:
structure_avoid
The yellow dots are the special structure avoidance. I guess the paths will look similar to the white ones. If I understand you correctly this is even more what you are trying to archive.

Why not have a gradient based on the range for special structures [SOURCE, MINERAL, SK_LAIR], which tries to avoid going close to these structures, but low enough to still use it if it makes sense.

PR #325, opened 2017-07-30
Fix and improve getEnergyFromHostileStructures

samogot:


src/role_nextroomer.js
```diff
@@ -217,25 +217,7 @@ roles.nextroomer.settle = function(creep) {
...
-        structure.destroy();
-        return true;
-      }
-      creep.say('ho: ' + structure.pos, true);
-      creep.log(structure.structureType);
-      creep.moveTo(structure);
-      creep.withdraw(structure, RESOURCE_ENERGY);
-      return true;
-    }
+    return creep.getEnergyFromHostileStructures();
```
TooAngel: I think this is anyway included in `Creep.getEnergy`, so I'm not sure if it is needed here.

And if, I guess if should be 
```
if (creep.getEnergyFromHostileStructures()) {
  return true;
}
```
So that the rest is executed if `getEnergyFromHostileStructures` fails.

src/prototype_room_utils.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+Room.structureHasEnergy = structure => structure.store && structure.store.energy || structure.energy;
+
+Room.structureIsEmpty = structure => (!structure.store || _.sum(structure.store) === 0) && !structure.energy && !structure.mineralAmount && !structure.ghodium && !structure.power;
```
TooAngel: I like, but is it used somewhere?

samogot: in `getEnergyFromHostileStructures` and in #326

src/role_nextroomer.js
samogot: fixed

democratic-collaboration:
A new review, yeah.

    Votes: 245/429
    Coefficient: 0.571095571096
    Merging in 3 days 9 hours
    Age 0 days 21 hours

PR #326, opened 2017-07-30
Allow miss-placed hostile storage with huge amount of energy

samogot:
Depends on #325

### DIFF
--- a/src/prototype_room_basebuilder.js
+++ b/src/prototype_room_basebuilder.js
@@ -67,7 +67,7 @@ Room.prototype.destroyStructure = function(structure) {
     structuresMin = 1;
   }
 
-  if (structures.length > structuresMin) {
+  if (structures.length > structuresMin && (structure.my || Room.structureIsEmpty(structure))) {
     this.log('Destroying: ' + structure.structureType + ' ' + JSON.stringify(structure.pos));
     structure.destroy();
     return true;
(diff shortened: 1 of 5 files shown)


src/prototype_room_basebuilder.js
```diff
@@ -67,7 +67,7 @@ Room.prototype.destroyStructure = function(structure) {
     structuresMin = 1;
   }
 
-  if (structures.length > structuresMin) {
+  if (structures.length > structuresMin && Room.structureIsEmpty(structure)) {
```
TooAngel: I think this breaks the case where the room layout changes and structures need to be placed somewhere else.

Hostile structures get removed after some time, owned structures can keep their resources and will be never replaced.

To address your case you could do `&& (structure.my || Room.structureIsEmpty(structure))` or something more sophisticated.

samogot: Yeah, that's why I was asking about should I make this PR at all. This changes are will be very rare-used, may break something else, and any way, it's better if such cases would be monitored manualy) So I'm not insist this to be merged

TooAngel: Overall I like the changes and I think it is fine to merge. Just this line I'm pretty sure this will break the cases where the room layout is changed and structures need to be moved to somewhere else.

samogot: fixed

democratic-collaboration:
A new review, yeah.

    Votes: 245/429
    Coefficient: 0.571095571096
    Merging in 3 days 9 hours
    Age 0 days 21 hours

PR #327, opened 2017-07-30
Fix carry spawn rate for low RCL

samogot:
Carries is spawned faster then they can move to sourcer to change dropped resource amount

### DIFF
--- a/src/prototype_creep_harvest.js
+++ b/src/prototype_creep_harvest.js
@@ -85,5 +85,5 @@ Creep.prototype.spawnCarry = function() {
   if (resourceAtPosition > carryCapacity) {
     Game.rooms[this.memory.base].checkRoleToSpawn('carry', 0, this.memory.routing.targetId, this.memory.routing.targetRoom);
   }
-  this.memory.wait = waitTime;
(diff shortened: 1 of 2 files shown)


src/prototype_creep_harvest.js
```diff
@@ -85,5 +85,5 @@ Creep.prototype.spawnCarry = function() {
   if (resourceAtPosition > carryCapacity) {
     Game.rooms[this.memory.base].checkRoleToSpawn('carry', 0, this.memory.routing.targetId, this.memory.routing.targetRoom);
   }
-  this.memory.wait = waitTime;
+  this.memory.wait = Math.max(waitTime, 50);
```
TooAngel: I would improve the `wait` time via a PID controller, we have the potential, the differential would be next.

As first step, we could have a fixed value, but this should be in the `config.js`

samogot: fixed

democratic-collaboration:
A new review, yeah.

    Votes: 245/429
    Coefficient: 0.571095571096
    Merging in 3 days 9 hours
    Age 0 days 21 hours

PR #337, opened 2017-08-06
Rewrite visualizer using native RoomVisual

samogot:
- add functions for visualizing costmatrix and PathFinder search results (but not called from anywhere)

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -137,6 +137,7 @@
     "OBSTACLE_OBJECT_TYPES": false,
     "CARRY_CAPACITY": false,
     "HARVEST_POWER": false,
+    "RoomVisual": false,
     "LOOK_RESOURCES": false
   },
   "node": true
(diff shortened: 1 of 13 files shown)


democratic-collaboration:
A new review, yeah.

    Votes: 257/441
    Coefficient: 0.582766439909
    Merging in 3 days 5 hours
    Age 0 days 22 hours

samogot:
improved

TooAngel:
I think the visualizer shouldn't be handled via config variables. I think manual `utils` commands make more sense. Checking a room via visualizer is easier by executing `utils.showStructures(ROOMNAME)` or `Game.rooms.A1B2.showStructures()`, then by changing the config and redeploy.

samogot:
Nah, native visualizer draws only for one tick

TooAngel:
Ah, ok, didn't know, good to know :-)

Having commands instead of config and using memory to check if it should be drawn?

What are your experiences? I got the impression that it is annoying to use the config, especially because of the reset of the cache

samogot:
yeah, I'm annoyed too. We can move this config to Memory, and change memory values from console)

PR #339, opened 2017-08-06
Add new towerdrainer role

samogot:


src/role_towerdrainer.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  const restPos = creep.memory.restPosition;
+  if (!restPos) {
+    creep.log('no position');
+    creep.moveRandom();
+    return false;
+  }
+
```
TooAngel: You can execute `creep.heal(creep)` always.

src/role_towerdrainer.js
```diff
@@ -0,0 +1,73 @@
...
+  } else if (creep.pos.isEqualTo(restPos.x, restPos.y)) {
+    creep.move(creep.memory.attackDirrection);
+  } else {
+    creep.moveTo(restPos.x, restPos.y);
+  }
+
+  return true;
+};
+
+roles.towerdrainer.execute = function(creep) {
```
TooAngel: `.execute` is the old system. I think this can be executed within the creep execution layer. I'm usually adding it to see when it is executed. Just as information.

src/role_towerdrainer.js
```diff
@@ -0,0 +1,73 @@
...
+    return false;
+  }
+
+  if (creep.pos.roomName === attackRoom || creep.pos.isBorder(-1) && creep.pos.isNearTo(restPos.x, restPos.y)) {
+    creep.heal(creep);
+    creep.move(creep.memory.restDirrection);
+  } else if (creep.hits < creep.hitsMax) {
+    creep.heal(creep);
+  } else if (creep.pos.isEqualTo(restPos.x, restPos.y)) {
+    creep.move(creep.memory.attackDirrection);
```
TooAngel: When you arrive in a new room, your health is 100%, the tower attacks you. You will loose hit points.
If you heal preemptive you can reduce that. Which would also be fixed when healing every tick.

src/role_towerdrainer.js
samogot: I can. but it is additional 0.2 cpu for no reason

src/role_towerdrainer.js
samogot: I already heal when I in attackRoom. It's `creep.pos.roomName === attackRoom` if-branch. I do not need to heal on tick when i'm mooving from rest room into target room, as I won't get any damage (from tower) in that tick

src/role_towerdrainer.js
samogot: ok) I just copied it from somewhere)

samogot:
fixed

democratic-collaboration:
A new review, yeah.

    Votes: 257/441
    Coefficient: 0.582766439909
    Merging in 6 days 5 hours
    Age 0 days 0 hours

PR #344, opened 2017-08-08
Safe way to invalidate room memory

samogot:
Global variables are tied to server node. Due jumping between nodes, it is hard to invalidate global cache from console without reruning code. Simple clearing cache from console may cause errors and even incorrect layout.


democratic-collaboration:
A new review, yeah.

    Votes: 257/441
    Coefficient: 0.582766439909
    Merging in 4 days 3 hours
    Age 0 days 0 hours

PR #362, opened 2017-08-16
updated jshint added "unused": true, fixed jshint issues

mschultheiss83:


mschultheiss83:
close here, all in one commit see #363

PR #375, opened 2017-08-28
Improve mineral handling

samogot:
- send more resources (bought and intermediate reagents) between rooms
- fix terminal overflowed with energy
- config option for debug logging
- stop reaction if no more resources left
- early suicide mineral creep to not waste dropped minerals
- count all tiers of reaction to check if we reach the limit

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -57,7 +57,8 @@ global.config = {
   debug: {
     getPartsConfLogs: false,
     queue: false,
-    spawn: false
+    spawn: false,
+    mineral: false
   },
 
   tower: {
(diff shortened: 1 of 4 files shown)


.jshintrc
```diff
@@ -139,6 +139,7 @@
     "CARRY_CAPACITY": false,
     "HARVEST_POWER": false,
     "RoomVisual": false,
+    "RESOURCE_HYDROXIDE",
```
stu43005: ``"RESOURCE_HYDROXIDE": false,``

PR #383, opened 2017-09-05
Fix fresh startup

TooAngel:
This is surely not the best solution, but currently the startup is broken as happened on #botarean


democratic-collaboration:
A new review, yeah.

    Votes: 290/484
    Coefficient: 0.599173553719
    Merging in 3 days 23 hours
    Age 0 days 0 hours

PR #385, opened 2017-09-07
[WIP] efficient scouts

Nevrdid:
-scouts now prioritize not recently seen rooms  …
-scout no more stuck on borders
-increased number of scouts with RCL

TODO : move levels from creeps memory to room memory and iterate along it at scout spawn before create his own levels which will be lighter

example of behavior around https://screeps.com/a/#!/room/shard1/W33N34 if nothing bug. ^^ Looks well right now.

### DIFF
--- a/src/role_scout.js
+++ b/src/role_scout.js
@@ -54,6 +54,9 @@ roles.scout.action = function(creep) {
   } else if (creep.memory.search) {
     roles.scout.move(creep);
   }
+  if (creep.isStuck()) {
+    roles.scout.unStuckIt(creep);
+  }
 };
 
 const isSafe = function(roomMem) {
@@ -157,10 +160,6 @@ roles.scout.move = function(creep) {
       }
     }
     let returnCode = creep.move(creep.pos.getDirectionTo(search.path[0]));
-  }
-
-  if (creep.isStuck()) {
-    roles.scout.unStuckIt(creep);
   }
 };
 
--- a/src/role_scout.js
+++ b/src/role_scout.js
@@ -18,17 +18,12 @@ roles.scout.execute = function(creep) {
     creep.memory.skip = [];
   }
 
-  if (!creep.memory.notifDisabled) {
+  if (!creep.memory.notifyDisabled) {
     creep.notifyWhenAttacked(false);
-    creep.memory.notifDisabled = true;
+    creep.memory.notifyDisabled = true;
   }
 
-  return roles.scout.breadthFirstSearch(creep);
-};
-
-roles.scout.haveNotSeen = function(creep, room) {
-  return creep.memory.search.seen.indexOf(room) === -1 &&
-    creep.memory.skip.indexOf(room) === -1;
+  return roles.scout.search(creep);
 };
 
 roles.scout.unStuckIt = function(creep, targetPos) {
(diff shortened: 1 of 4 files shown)


src/role_scout.js
```diff
@@ -13,153 +13,219 @@ roles.scout.settings = {
...
-        ignoreCreeps: true,
-        costCallback: creep.room.getCostMatrixCallback()
-      });
-      //}
-      return true;
     }
-    creep.say(creep.pos.getDirectionTo(search.path[0]));
-    let returnCode = creep.move(creep.pos.getDirectionTo(search.path[0]));
-  };
+    creep.moveTo(25, 25);
```
lurr: The center may be obstructed- perhaps a different method is in order.

Nevrdid: I supose something as creep.moveTo(25, 25, {maxOps: 100}) should avoid to lost the cpu in that case.

src/role_scout.js
```diff
@@ -13,153 +13,226 @@ roles.scout.settings = {
   maxLayoutAmount: 1
 };
 
-function onBorder(creep) {
+roles.scout.execute = function(creep) {
+  if (creep.memory.skip === undefined) {
+    creep.memory.skip = [];
+  }
+
+  creep.notifyWhenAttacked(false);
```
samogot: This always costs at least 0.2 cpu. Maybe we can invoke it only once per creep?

src/role_scout.js
```diff
@@ -13,153 +13,226 @@ roles.scout.settings = {
...
-      //       creep.log(creep.pos + ' ' + targetPosObject + ' ' + JSON.stringify(search));
-      if (creep.isStuck() && onBorder(creep)) {
-        creep.say('imstuck at the border', true);
-        if (config.room.scoutSkipWhenStuck) {
-          creep.say('skipping', true);
-          creep.memory.scoutSkip = true;
-          delete creep.memory.last; // Delete to reset stuckness.
-        }
+  if (search.path.length === 0 || (creep.inBase() && creep.room.memory.misplacedSpawn)) {
+    if (creep.isStuck() && roles.scout.onBorder(creep)) {
```
samogot: `creep.pos.isBorder()`?

src/role_scout.js
```diff
@@ -13,153 +13,226 @@ roles.scout.settings = {
...
     }
+    return false;
+  };
 
-    if (creep.isStuck()) {
-      creep.moveRandom();
-      creep.say('ImStuck', true);
-      creep.log('Scout Stuck, Randomly Moving: ' + JSON.stringify(creep.memory.last) + ' ' + JSON.stringify(creep.isStuck()));
-      return true;
+  let search = PathFinder.search(
```
samogot: what about caching path in memory? at least for several ticks, like `moveTo`. Or maybe encapsulate search inside `moveToMy` somehow, and make cache layer there later.

src/role_scout.js
```diff
@@ -13,153 +13,219 @@ roles.scout.settings = {
...
+
+  creep.targetPosObject = new RoomPosition(25, 25, creep.memory.search.target);
+
+  if (creep.room.getEnemys().length) {
+    creep.say('Afraid', true);
+    creep.memory.scoutSkip = true;
+    return false;
+  }
+  if (creep.memory.last && creep.memory.last.pos3 && creep.room.name !== creep.memory.last.pos3.roomName) {
+    let roomMem = Memory.rooms[creep.room.name];
```
samogot: `creep.room.memory`

src/role_scout.js
Nevrdid: was previously here but can add config check with a flag in memory for do it only once.
maybe better in another PR

src/role_scout.js
```diff
@@ -13,153 +13,219 @@ roles.scout.settings = {
...
+  (creep.memory.scoutSkip || creep.room.name === creep.memory.search.target)
+  && roles.scout.switchTarget(creep);
+
+  if (!creep.memory.search.target) {
+    creep.log('Suiciding: ' + JSON.stringify(creep.memory.search));
+    creep.suicide();
+    return true;
+  }
+
+  creep.targetPosObject = new RoomPosition(25, 25, creep.memory.search.target);
```
samogot: pass this as separate parameter of `unStuckIt` and `move` functions would be much more clear then set custom property of game object

src/role_scout.js
Nevrdid: as pathing layout don't use a ratio 2:1 as plain/swamp we can't use moveToMy as it is actually.
but we can for sure memorize path by creep.
already thought about memorize rooms deep, which is memorized in scout, into base room but as it is working actually, i thought about do it in next PR.

src/role_scout.js
Nevrdid: yup should be better ^^ onBorder was already here as i didn't look if there was a creep proto.

src/role_scout.js
Nevrdid: yup, ^^' was checking if room was visible on previous version and didn't switch as room always exist here

src/role_scout.js
```diff
@@ -13,153 +13,219 @@ roles.scout.settings = {
...
-function onBorder(creep) {
-  return creep.pos.x === 49 || creep.pos.x === 0 ||
-    creep.pos.y === 49 || creep.pos.y === 0;
-}
+roles.scout.execute = function(creep) {
+  if (creep.memory.skip === undefined) {
+    creep.memory.skip = [];
+  }
+
+  if (!creep.memory.notifDisabled) {
```
samogot: typo: notif**y**Disabled

src/role_scout.js
```diff
@@ -13,153 +13,219 @@ roles.scout.settings = {
...
+
+  let targetPos = new RoomPosition(25, 25, creep.memory.search.target);
+
+  if (creep.room.getEnemys().length) {
+    creep.say('Afraid', true);
+    creep.memory.scoutSkip = true;
+    return false;
+  }
+  if (creep.memory.last && creep.memory.last.pos3 && creep.room.name !== creep.memory.last.pos3.roomName) {
+    let roomMem = creep.room.memory;
```
samogot: just to be clear, creep.room.memory may also be undefined if we see this room for first time. Not sure if you initialize it somewhere else)

src/role_scout.js
Nevrdid: ✍️

src/role_scout.js
Nevrdid: yeah... will verify ^^'

src/role_scout.js
TooAngel: I usually use `PathFinder` with a range of ~20

src/role_scout.js
kahojyun: I'm also considering caching path in memory. Currently, when scout move from room A to room B, room A may become invisible, and `PathFinder` gives different search result because  `getCostMatrixCallback` returns a different `costMatrix` for room A. As a result, scout frequently gets stuck on borders. There are two solutions:
1. Cache path in memory, thus scout won't be influenced by different path search results.
2. Cache another version of `costMatrix` in memory for scout.
I prefer the first solution because it's not hard to be implented, and can benefit all roles that use `pathFinder`.

TooAngel: ^^ That is a really good finding. Had that issue a couple of days ago, couldn't find the reason.

I like the 2., too. Caching the complete path would result in less CPU usage.

src/role_scout.js
Nevrdid: as rooms are handle before creeps in, the memory is already filled with state and lastSeen at least

src/role_scout.js
Nevrdid: the issue is definitly fixed by  `if (!Memory.rooms[roomName] || (Memory.rooms[roomName].lastSeen < (Game.time - 5))) {` before use costatrix.
the issue caching path/Cm is that depending the number of rooms / scouts by base rooms, it will surely quick go to a leack of memory. already seen problems with the size of search.levels while using multiple scouts.

Nevrdid:
Updated, reviews are welcome :) i think to include it before merge as substitute to actual one for allow people to choose and do comparaison between both with grafana.
Some more upgrade i'm working on locally but breaking some things ^^'.
Actually trying to search bad rooms patern for it.

samogot:
Is it still WIP? should I re-review it before or after conflict resolving?)

Nevrdid:
hum i'm not sure. Pretty sure this version work as i pushed it but i may i find issues i solved on my workbracn. i should check and fetch if needed. But the behave stay the same. I supose you can already comment anything about logic but syntax and structure may change.

src/role_scout.js
```diff
@@ -10,156 +10,195 @@ roles.scout = {};
...
-    if (!creep.memory.search.target) {
-      creep.log('Suiciding: ' + JSON.stringify(creep.memory.search));
-      creep.suicide();
-      return true;
-    }
-    let targetPosObject = new RoomPosition(25, 25, creep.memory.search.target);
+roles.scout.action = function(creep) {
+  if (!creep.memory.search) {
+    roles.scout.initSearch(creep);
+    creep.say('i-' + creep.memory.search.startRoom, true);
```
samogot: why public?

samogot:
any progress here?

PR #387, opened 2017-09-07
[WIP] Reservers fix

Nevrdid:


src/role_reserver.js
```diff
@@ -41,6 +41,8 @@ roles.reserver.action = function(creep) {
     }
   }
 
+  creep.say('Reserver -> ' + creep.memory.routing.targetRoom, true);
```
lurr: Why this? It just fills up more, and it gives information publicly.
This would be fine imo if it was not public

src/prototype_room_external.js
```diff
@@ -251,14 +251,14 @@ Room.prototype.checkAndSpawnReserver = function() {
     }
   };
   // TODO move the creep check from the reserver to here and spawn only sourcer (or one part reserver) when controller.level < 4
+  /** TODO look if it bug. if yeah, cf garbage solution (Nevrdid)
```
samogot: `/**` is JSDoc comment. use `/*` for simple block comments

src/role_reserver.js
```diff
@@ -41,10 +41,29 @@ roles.reserver.action = function(creep) {
     }
   }
 
+  creep.say('Reserver -> ' + creep.memory.routing.targetRoom);
+
   // TODO this should be enabled, because the reserver should flee without being attacked
   creep.notifyWhenAttacked(false);
```
samogot: As well, you didn't touch it, but it is better to call it once

TooAngel:
@lurr `democratic collaboration` works via pull request reviews. Better `Request changes` in the review process then `:thumbsdown`

src/role_reserver.js
TooAngel: Not necessary here.

IMHO we should prefer public over private, maybe `config` variable to disable public saying.
The reason for public is, that IMHO the TooAngel AI should be more seen as an NPC with a high level of transparency. Also the TooAngel users can debug others rooms, too.

src/prototype_creep_resources.js
```diff
@@ -645,42 +645,34 @@ Creep.prototype.transferEnergyMy = function() {
 
 Creep.prototype.reserverSetLevel = function() {
   this.memory.level = 2;
-  if (this.room.controller.reservation && this.room.controller.reservation.ticksToEnd > 4500) {
+  if (this.room.controller.reservation && (Game.rooms[this.memory.base].energyCapacityAvailable < 1300 || this.room.controller.reservation.ticksToEnd > 4500)) {
```
TooAngel: Both number are better. Either calculated via the `BODYPART_COST` constants, or from `config`

src/prototype_creep_resources.js
```diff
@@ -645,42 +645,34 @@ Creep.prototype.transferEnergyMy = function() {
...
     this.memory.level = 1;
   }
   if (!this.room.controller.my && this.room.controller.reservation && this.room.controller.reservation.username !== Memory.username) {
     this.memory.level = 5;
   }
 };
 
-let callStructurer = function(creep) {
-  const structurers = creep.room.findPropertyFilter(FIND_MY_CREEPS, 'memory.role', ['structurer']);
+Creep.prototype.callStructurer = function() {
```
TooAngel: Fine.

Even better as `Room.prototype` and search for the closest room.

src/prototype_creep_resources.js
```diff
@@ -645,42 +645,34 @@ Creep.prototype.transferEnergyMy = function() {
...
-  if (creep.inBase()) {
-    return false;
-  }
-
-  if (!Game.rooms[creep.memory.base].storage) {
-    return false;
-  }
-
-  if (!creep.room.exectueEveryTicks(1000)) {
+Creep.prototype.callCleaner = function() {
```
TooAngel: Fine.

Even better as `Room.prototype` and search for the closest room.

src/prototype_creep_resources.js
```diff
@@ -694,14 +686,14 @@ let checkSourcerMatch = function(sourcers, source_id) {
   return false;
 };
 
-let checkSourcer = function(creep) {
-  const sources = creep.room.find(FIND_SOURCES);
-  const sourcers = creep.room.findPropertyFilter(FIND_MY_CREEPS, 'memory.role', ['sourcer']);
+Creep.prototype.checkSourcer = function() {
```
TooAngel: This should be definitely as `Room.prototype` and search for the closest room. These checks should happen in the room and not on a creep.

samogot:
I'm not very interested in this PR as I do not understand the problem you were fixing... but is it still WIP? or rather, do we still need it, and will you continue work on it?)

Nevrdid:
started fixing reserver and progressivly modify other things about it in my workbranch then it finished here. I should for sure close this one as there are multiple things without link to reserver. Still some things i would see merged as allow remote mining without control, level merged to reserver settings, prerequisites changed for reservation and other things.. But i supose it have more chance to appear in another PR ^^'

PR #388, opened 2017-09-07
slower upgrader speach

Nevrdid:


lurr:
There's a maximum limit to how big a word can be when passed to creep.say()
Along with this, it will repeat itself _ten times_ in a row.

Personally, I'd prefer to keep the current system that's in place here. I may not use this code, but I'm putting in my word.

Nevrdid:
true, forget about the limit but the main target was to allow us to read it on fast ticks rate as actually on botarena.
I'll separate again name & value & add a config value for setup how long to keep each word.

democratic-collaboration:
A new review, yeah.

    Votes: 117/485
    Coefficient: 0.241237113402
    Merging in 15 days 2 hours
    Age 0 days 1 hours

democratic-collaboration:
A new review, yeah.

    Votes: 122/485
    Coefficient: 0.251546391753
    Merging in 7 days 11 hours
    Age 0 days 0 hours

TooAngel:
IMHO we should remove the shouting of the `upgrader` creep.
There was a couple of negative feedback on this. If it should stay I would phrase it to a positive list, based on the reputation.
Long term the info can be moved to a public memory segment.

Nevrdid:
agree, will try to do something more positive so ^^ WIP it until i have something good

PR #390, opened 2017-09-07
[WIP]allow harvesters transfer to other creeps

Nevrdid:


src/role_harvester.js
```diff
@@ -102,6 +102,10 @@ roles.harvester.preMove = function(creep, directions) {
   if (directions && creep.memory.routing.reverse) {
     directions.direction = directions.backwardDirection;
   }
+  if (directions.backwardDirection && reverse && roles.scout.isStuck(creep)) {
```
TooAngel: The `isStuck` method should be on the creep, not on the scout.

Nevrdid: 👍

Nevrdid: 👍 yeah, was firstly made for period when harvesters are multiples in rooms and stuck themself often as there are harvester full of energy which wanna go to construction site near the controller and also other carrys trying to gonna fill extensions near container whitch are empty. not even sure it's the better solution now x) i'll WIP it as i'll not can work alot on next days.

Nevrdid:
👍 yeah, was firstly made for period when harvesters are multiples in rooms and stuck themself often as there are harvester full of energy which wanna go to construction site near the controller and also other carrys trying to gonna fill extensions near container whitch are empty. not even sure it's the better solution now x) i'll WIP it as i'll not can work alot on next days.

PR #393, opened 2017-09-09
bug fix

kahojyun:
- don't remove border walls placed by system
- repairer stuck when there are system walls in room
- Memory initialize
- fix carry withdrawing energy from base storage
- carry helper shouldn't transfer to creeps in targetRoom

### DIFF
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -48,6 +48,12 @@ Creep.prototype.checkEnergyTransfer = function(otherCreep) {
   return offset + _.sum(this.carry) > carryPercentage * this.carryCapacity;
 };
 
+// return true if helper can't transfer to creep
+Creep.prototype.checkHelperNoTransfer = function(creep) {
+  return this.memory.helper && !creep.memory.helper &&
+        creep.memory.role === 'carry' && creep.memory.base !== this.memory.base;
+};
+
 Creep.prototype.findCreepWhichCanTransfer = function(creeps) {
   for (let i = 0; i < creeps.length; i++) {
     const otherCreep = creeps[i];
@@ -55,7 +61,10 @@ Creep.prototype.findCreepWhichCanTransfer = function(creeps) {
       continue;
     }
 
-    if (Game.creeps[otherCreep.name].memory.role === 'carry') {
+    if (otherCreep.memory.role === 'carry') {
+      if (otherCreep.checkHelperNoTransfer(this)) {
+        continue;
+      }
       return this.checkEnergyTransfer(otherCreep);
     }
     continue;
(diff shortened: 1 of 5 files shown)


democratic-collaboration:
A new review, yeah.

    Votes: 60/487
    Coefficient: 0.123203285421
    Merging in 8 days 11 hours
    Age 0 days 6 hours

democratic-collaboration:
A new review, yeah.

    Votes: 65/487
    Coefficient: 0.133470225873
    Merging in 7 days 16 hours
    Age 0 days 23 hours

democratic-collaboration:
A new review, yeah.

    Votes: 124/488
    Coefficient: 0.254098360656
    Merging in 6 days 7 hours
    Age 1 days 3 hours

src/prototype_creep_resources.js
```diff
@@ -297,7 +303,8 @@ Creep.prototype.transferToCreep = function(direction) {
   const creeps = adjacentPos.lookFor('creep');
   for (let i = 0; i < creeps.length; i++) {
     const otherCreep = creeps[i];
-    if (!checkCreepForTransfer(otherCreep)) {
+    // don't transfer to other base's creep, fix carry helper
+    if (!checkCreepForTransfer(otherCreep) || otherCreep.memory.base !== this.memory.base) {
```
TooAngel: I don't think this is the best approach. Maybe fixes the current issue.

I had a couple of cases where the external rooms of different base rooms are crossing each other and it doesn't matter to which room the energy is transferred.

I think the better approach would be checking for the direction. `helper` is the newest sort of carry usage, maybe we should rethink that, add a value to the memory to be able to check for them, have a different role type, ...

kahojyun: What about `this.memory.helper && !otherCreep.memory.helper && otherCreep.memory.base !== this.memory.base`?

In fact, I disabled transfer to carry in PR #403. I don't think it necessary to allow energy transfer between carries because it will mess up carry's routing.

TooAngel: I think this approach is better. We can also restrict, that helpers can only transfer to helpers, carries only to carries that should be fine.

Not sure if I understand your last sentence

kahojyun: check out the new commit plz :)

I'm sorry for my poor English if my last sentence confuse you...  I found that carries near `pathStart` will get energy from other carries and not move to their own target.  This problem is obvious in low RCL rooms.  I just forbade energy transfer between carries in my scripts so that every carry can reach their targets.

democratic-collaboration:
A new review, yeah.

    Votes: 362/494
    Coefficient: 0.732793522267
    Merging in 3 days 18 hours
    Age 0 days 5 hours

TooAngel:
@kahojyun If you add 'Fixes #410' to the commit message of one of your commits, the issue will be closed automatically.

PR #403, opened 2017-09-10
[WIP] improve carry recycling

kahojyun:
The main point of this commit is improving energy harvest efficiency.

1. Improved the way to check extra carries.  Carries will be recycled if they are waiting for sourcer.
2. Before recycling carries with a spawn, check spawn queue and assign a new target to carry.
3. Sourcers calculate spawn rate depending on `resourceAtPostion`.
4. Added an option to switch on/off energy transfer between carries.  It's turned off by default.
5. Modified spawn priority for reserver and carry.

With this improvement, carries will work more efficiently.  Tested on the official server:

Stored about 900k energy in storage before upgrading to RCL6. (five sources in extern room)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -142,10 +142,13 @@ global.config = {
       1800: [8, 15], // RCL 5
       2300: [11, 21], // RCL 6
     },
+    transferToCarry: false,
+    // limit carry spawn rate when RCL < 4
+    minSpawnRate: 50,
     // when ticksToLive > recycleThreshold, reuse carry
     recycleThreshold: 200,
-    transferToCarry: false,
-    minSpawnRate: 50,
+    // keep at least one carry for a target
+    ensureOne: true,
     // Percentage should increase from base to target room. Decrease may cause stack on border
     carryPercentageBase: 0.1,
     carryPercentageHighway: 0.2,
--- a/src/config.js
+++ b/src/config.js
@@ -142,6 +142,9 @@ global.config = {
       1800: [8, 15], // RCL 5
       2300: [11, 21], // RCL 6
     },
+    // when ticksToLive > recycleThreshold, reuse carry
+    recycleThreshold: 200,
+    transferToCarry: false,
     minSpawnRate: 50,
     // Percentage should increase from base to target room. Decrease may cause stack on border
     carryPercentageBase: 0.1,
@@ -243,9 +246,9 @@ global.config = {
       defender: 12,
       defendranged: 13,
       nextroomer: 15,
-      carry: 17,
+      reserver: 17,
       sourcer: 18,
-      reserver: 19,
+      carry: 19,
     },
   },
 };
--- a/src/prototype_creep.js
+++ b/src/prototype_creep.js
@@ -63,7 +63,7 @@ Creep.prototype.handle = function() {
     return;
   }
 
-  if (this.memory.recycle && this.memory.moveToSpawn) {
+  if (this.memory.recycle) {
     Creep.recycleCreep(this);
     return;
   }
--- a/src/prototype_creep.js
+++ b/src/prototype_creep.js
@@ -63,7 +63,7 @@ Creep.prototype.handle = function() {
     return;
   }
 
-  if (this.memory.recycle) {
+  if (this.memory.recycle && this.memory.moveToSpawn) {
     Creep.recycleCreep(this);
     return;
   }
--- a/src/prototype_creep_harvest.js
+++ b/src/prototype_creep_harvest.js
@@ -67,18 +67,13 @@ Creep.prototype.spawnCarry = function() {
     Math.sqrt(carryCapacity / Math.max(resourceAtPosition, carryCapacity));
 
   if (resourceAtPosition > carryCapacity) {
-    baseRoom.checkRoleToSpawn('carry', 0, this.memory.routing.targetId,
-      this.memory.routing.targetRoom, undefined, undefined, {
-        checkRecycle: true,
-      });
+    baseRoom.checkRoleToSpawn('carry', 0, this.memory.routing.targetId, this.memory.routing.targetRoom);
   }
 
-  // low minSpawnRate helps carry recycling
-  let minSpawnRate = 10;
-
-  // higher minSpawnRate when RCL < 4
+  // limit carry spawn rate when RCL < 4
   if (baseRoom.controller.level < 4) {
-    minSpawnRate = config.carry.minSpawnRate;
+    this.memory.wait = Math.max(waitTime, config.carry.minSpawnRate);
+  } else {
+    this.memory.wait = waitTime;
   }
-  this.memory.wait = Math.max(waitTime, minSpawnRate);
 };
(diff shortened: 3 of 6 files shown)


src/prototype_creep_harvest.js
```diff
@@ -63,15 +63,22 @@ Creep.prototype.spawnCarry = function() {
     resourceAtPosition += _.sum(container.store);
   }
 
+  const waitTime = carryCapacity / (HARVEST_POWER * workParts) *
+    Math.sqrt(carryCapacity / Math.max(resourceAtPosition, carryCapacity));
```
samogot: like it

src/prototype_creep_routing.js
```diff
@@ -267,6 +267,10 @@ Creep.prototype.moveByPathMy = function(route, routePos, start, target, skipPreM
         if (this.memory.killPrevious) {
           this.killPrevious();
         }
+        if (this.memory.checkRecycle && this.isStuck() && !this.memory.routing.reverse) {
```
samogot: I do not like it a lot. carry may stuck not only because of lack of resources at sourcer
either way, name `recycleOnStuck` would be better then `checkRecycle`

src/prototype_creep.js
```diff
@@ -63,7 +63,7 @@ Creep.prototype.handle = function() {
     return;
   }
 
-  if (this.memory.recycle) {
+  if (this.memory.recycle && this.memory.moveToSpawn) {
```
samogot: do not like it at all. `creep.memory.recycle` meant to be universal and easily to set manually from interface, if we need it

src/role_carry.js
```diff
@@ -163,6 +184,27 @@ roles.carry.preMove = function(creep, directions) {
...
+      if (creep.ticksToLive > config.carry.recycleThreshold && roles.carry.reset(creep)) {
+        creep.memory.recycle = false;
+      } else {
+        const spawn = creep.pos.findClosestByRangePropertyFilter(FIND_MY_STRUCTURES, 'structureType', [STRUCTURE_SPAWN]);
+        if (spawn && spawn.recycleCreep(creep) === 'OK') {
+          return true;
+        }
+      }
+      if (roles.carry.checkOnPathStart(creep)) {
+        creep.memory.moveToSpawn = true;
```
samogot: it's better to handle carries pre-recycle logic with another memory flag, and set recycle flag here

src/prototype_creep_routing.js
samogot: maybe we may move it to carries pre-move?

src/role_carry.js
kahojyun: you're right, I'll use `resetTarget` for pre-recycle logic

src/prototype_creep_harvest.js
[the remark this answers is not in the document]
kahojyun: Reduce `waitTime` according to `resourceAtPosition`.  I use `Math.sqrt` in the hope that it will reach an equilibrium quickly (not call/spawn too much carries).

src/prototype_creep_routing.js
kahojyun: It's a little bit tricky to move this to pre-move because I need to use `path.length`.  Maybe we can store `path.length` in creeps' memory.

samogot: we can pass it the same way as directions

src/config.js
```diff
@@ -139,7 +139,13 @@ global.config = {
       1800: [8, 15], // RCL 5
       2300: [11, 21], // RCL 6
     },
+    transferToCarry: false,
+    // limit carry spawn rate when RCL < 4
     minSpawnRate: 50,
+    // when ticksToLive > recycleThreshold, reuse carry
+    recycleThreshold: 200,
+    // keep at least one carry for a target
+    ensureOne: true,
```
kahojyun: I couldn't think up a better name...

src/prototype_creep_harvest.js
```diff
@@ -63,15 +63,17 @@ Creep.prototype.spawnCarry = function() {
...
-      nearCarries[0].memory.recycle = true;
-    }
+    baseRoom.checkRoleToSpawn('carry', 0, this.memory.routing.targetId, this.memory.routing.targetRoom);
+  }
+
+  // limit carry spawn rate when RCL < 4
+  if (baseRoom.controller.level < 4) {
+    this.memory.wait = Math.max(waitTime, config.carry.minSpawnRate);
+  } else {
+    this.memory.wait = waitTime;
```
kahojyun: Normally, `waitTime` won't get too small because of `Math.sqrt`.

src/config.js
```diff
@@ -139,7 +139,13 @@ global.config = {
       1800: [8, 15], // RCL 5
       2300: [11, 21], // RCL 6
     },
+    transferToCarry: false,
```
TooAngel: Not sure what the intention is.

The default behavior should be that carries transfer to carries. If I understand the implementation correct, that would be disabled here.
And there are two methods, something like `transferToCreep` which actually transferrs and something like `checkForTransfer` which lets the carry move in the other direction because it will get energy in this tick.

Both needs to be adapted.

src/role_carry.js
```diff
@@ -164,6 +192,35 @@ roles.carry.preMove = function(creep, directions) {
   }
 
   reverse = creep.pickupWhileMoving(reverse);
+
+  // recycle carry
```
TooAngel: I do not like the recycling idea sooo much.

If there is a problem with too many carries, the `call carry` logic should be improved.
But fine until the `call carry` logic is better.

kahojyun: The ultimate solution to `call carry` logic would be #294.  My solution is making use of 'negative feedback'.  The latter one is not as efficient as the former one, but it's easier to be implemented, and, in my opinion, their performance difference is insignificant.

src/config.js
TooAngel: I think it is fine, maybe `ensureOnePerTarget` ... :-)

src/prototype_creep_harvest.js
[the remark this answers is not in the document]
TooAngel: K, cool, let's see

src/prototype_creep_resources.js
```diff
@@ -106,7 +99,7 @@ Creep.prototype.pickupWhileMoving = function(reverse) {
   if (resources.length > 0) {
     const resource = resources[0];
     const amount = this.pickupOrWithdrawFromSourcer(resource);
-    return _.sum(this.carry) + amount > 0.5 * this.carryCapacity;
+    return this.checkEnergyTransfer(amount);
```
TooAngel: :+1:

src/role_carry.js
TooAngel: I'm fine with adding this one and I'm also not so sure about the perfect solution, especially due to the fact that paths should be shared over multiple sources, which makes it more tricky.

src/config.js
kahojyun: Maybe we can swap carries' target when they transfer to each other.

Nevrdid:
might have a look at my https://screeps.com/a/#!/room/shard1/W31N49 room. got a pretty good result with my amount calculation and creeps modifs.

PR #405, opened 2017-09-12
fix missing check in squad manager

Nevrdid:


kahojyun:
function `brain.cleanSquads` has the same issue. I think initializing memory (PR #393) can fix this.😄

Nevrdid:
thx, forget it ^^' close it so.

Issue #408, opened 2017-09-12
Creep Deadlock


nazzeth:
Hello, Newbie to this project/code and screeps in general. 

Noticing a lot of issues with path finding, did a bit of digging here in the issues section and saw that this has been looked at before: #224 

However, using a clone from just a few hours ago, I'm experiencing this issue. I find its generally one creep (carry) somewhere that is the issue, and having that creep suicide fixes all the traffic again, but it doesn't take long for the issue to re-occur. 

<img width="734" alt="nvidia share_2017-09-12_20-45-22" src="https://user-images.githubusercontent.com/6315368/30326778-34b99fae-97fc-11e7-9adc-9943f3853155.png">

Looking for any advice on how to resolve or workaround this issue :)

kahojyun:
plz try pr #409.  I am not sure if it can resolve this issue.

nazzeth:
Hi there kahojyun. 

I used the file, however, the issue continued to occur. 
<img width="424" alt="nvidia share_2017-09-13_01-28-02" src="https://user-images.githubusercontent.com/6315368/30339801-cd95e22a-9822-11e7-9e09-913175eaedc0.png">

Nevrdid:
may you try to use that ? : https://github.com/TooAngel/screeps/pull/406/files#diff-4c3abda147ce23a598ff4a6661aa5955R25

whith that : https://github.com/TooAngel/screeps/pull/406/files#diff-4c3abda147ce23a598ff4a6661aa5955R7

TooAngel:
Can you check if the creep is a helper? There should be a `helper` property in the creep memory. Maybe paste the compete creep memory here.

nazzeth:
@TooAngel - I don't see the helper property in the stuck creep's memory: 

![image](https://user-images.githubusercontent.com/6315368/30348620-9affebd8-9842-11e7-9b7a-76919e217929.png)

I'm also getting alot of these from harvesters that are stuck behind the stopped carry - 

```
[5:15:53 AM]555942 W8N3   harvester-7025       [room W8N3 pos 28,34] transferEnergyMy: Can not find target 70b371aae9daabd
[5:15:55 AM]555944 W8N3   harvester-7305       [room W8N3 pos 30,32] transferEnergyMy: Can not find target e3275c596dc7273
[5:15:55 AM]555944 W8N3   harvester-7025       [room W8N3 pos 27,35] transferEnergyMy: Can not find target e3275c596dc7273
[...]

PR #409, opened 2017-09-12
fix stuck carry

kahojyun:
fix issue #408

### DIFF
--- a/src/role_carry.js
+++ b/src/role_carry.js
@@ -71,11 +71,13 @@ roles.carry.handleMisplacedSpawn = function(creep) {
       const structure = creep.pos.findClosestByRange(FIND_MY_STRUCTURES, {
         filter: (object) => object.energy < object.energyCapacity,
       });
-      creep.moveTo(structure, {
-        ignoreCreeps: true,
-      });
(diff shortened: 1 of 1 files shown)


Nevrdid:
i did the same but with creep.isStuck() as ignore creeps

democratic-collaboration:
A new review, yeah.

    Votes: 238/494
    Coefficient: 0.481781376518
    Merging in 5 days 0 hours
    Age 0 days 3 hours

kahojyun:
Another bug found.  `source` is not a `RoomPosition` Object.

Nevrdid:
well play ^^ no more sure if my harvester PR is still necessary now. exept for fullfill spawn at rcl 1.

Nevrdid:
also, as with ignore creeps true it will disallow to have multiple harvesters mining the same time, it may be better not ignore them,

kahojyun:
In fact, the issue hasn't been resolved completely.  We shouldn't use `moveTo` there because Sourcers will block Carriers' path.  If we update the `costMatrix` dynamically instead of setting it to its final state in the beginning, we can use `moveToMy`.

PR #410, opened 2017-09-13
Fixes Exeception TypeError: if Memory.squads is undefined

mschultheiss83:
Exeception TypeError: Cannot convert undefined or null to object
Object.keys(Memory.squads) => Object.keys(undefined)


kahojyun:
Will be fixed by #393 :)

PR #412, opened 2017-09-15
Fixed carry stuck at `pathStart` when storage is full

kahojyun:
Carry will drop resources at `pathStart` if storage is full

Updated:
Fixed `transferToStructures`

### DIFF
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -360,7 +360,9 @@ const filterTransferrables = function(creep, object) {
     return false;
   }
 
-  if (object.structureType !== STRUCTURE_STORAGE && object.energy === object.energyCapacity) {
+  if (object.structureType === STRUCTURE_STORAGE ?
+    _.sum(object.store) + _.sum(creep.carry) > object.storeCapacity :
+    object.energy === object.energyCapacity) {
     return false;
   }
 
@@ -409,6 +411,12 @@ Creep.prototype.transferToStructures = function() {
       }
       transferred = this.transferAllResources(item.structure);
     }
+  }
+  if (transferred) {
+    return {
+      moreStructures: false,
+      transferred: transferred,
+    };
   }
   return false;
 };
--- a/src/role_carry.js
+++ b/src/role_carry.js
@@ -143,18 +143,20 @@ roles.carry.preMove = function(creep, directions) {
           return true;
         }
         reverse = creep.carry.energy - transferred.transferred > 0;
-      } else if (creep.memory.routing.pathPos === 0 && !(creep.room.storage && creep.room.storage.my && creep.room.storage.isActive())) {
+      } else if (creep.memory.routing.pathPos === 0) {
         creep.drop(RESOURCE_ENERGY);
         reverse = false;
-
-        let resourceAtPosition = 0;
-        const resources = creep.pos.lookFor(LOOK_RESOURCES);
-        for (const resource of resources) {
-          resourceAtPosition += resource.amount;
+        const storage = creep.room.storage;
+        if (!(storage && storage.my && storage.isActive())) {
+          let resourceAtPosition = 0;
+          const resources = creep.pos.lookFor(LOOK_RESOURCES);
+          for (const resource of resources) {
+            resourceAtPosition += resource.amount;
+          }
+          let amount = creep.room.getHarvesterAmount();
+          amount += Math.floor(resourceAtPosition / config.carry.callHarvesterPerResources);
+          creep.room.checkRoleToSpawn('harvester', amount, 'harvester');
         }
-        let amount = creep.room.getHarvesterAmount();
-        amount += Math.floor(resourceAtPosition / config.carry.callHarvesterPerResources);
-        creep.room.checkRoleToSpawn('harvester', amount, 'harvester');
       }
     }
     if (directions.backwardDirection && directions.backwardDirection !== null) {


src/role_carry.js
```diff
@@ -143,18 +143,20 @@ roles.carry.preMove = function(creep, directions) {
           return true;
         }
         reverse = creep.carry.energy - transferred.transferred > 0;
-      } else if (creep.memory.routing.pathPos === 0 && !(creep.room.storage && creep.room.storage.my && creep.room.storage.isActive())) {
+      } else if (creep.memory.routing.pathPos === 0) {
         creep.drop(RESOURCE_ENERGY);
```
Nevrdid: So will not creep drop energy anytime ? Are you sure the creep will succesfully transfer after that ? Should be placed l.151 i supose.

src/prototype_creep_resources.js
```diff
@@ -364,6 +364,10 @@ const filterTransferrables = function(creep, object) {
     return false;
   }
 
+  if (object.structureType === STRUCTURE_STORAGE && _.sum(object.store) + _.sum(creep.carry) > object.storeCapacity) {
```
Nevrdid: you should place it with l 363 for codeclimate: 
```
object.structureType === STRUCTURE_STORAGE ? 
_.sum(object.store) + _.sum(creep.carry) > object.storeCapacity :
object.energy === object.energyCapacity

```

src/role_carry.js
kahojyun: I have applied these changes to my scripts in the official server for several days.  Carry will drop energy only when `pathPos === 0 && inBase`.  I placed `drop` here because we shouldn't call harvester if we have storage.

src/prototype_creep_resources.js
kahojyun: thx

src/role_carry.js
Nevrdid: yeah but we should also not drop if we have storage not full

src/prototype_creep_resources.js
```diff
@@ -360,7 +360,9 @@ const filterTransferrables = function(creep, object) {
     return false;
   }
 
-  if (object.structureType !== STRUCTURE_STORAGE && object.energy === object.energyCapacity) {
+  if (object.structureType === STRUCTURE_STORAGE ?
```
Nevrdid: may also add STRUCTURE_TERMINAL && STRUCTURE_CONTAINER

src/role_carry.js
```diff
@@ -143,18 +143,20 @@ roles.carry.preMove = function(creep, directions) {
...
-          resourceAtPosition += resource.amount;
+        const storage = creep.room.storage;
+        if (!(storage && storage.my && storage.isActive())) {
+          let resourceAtPosition = 0;
+          const resources = creep.pos.lookFor(LOOK_RESOURCES);
+          for (const resource of resources) {
+            resourceAtPosition += resource.amount;
+          }
+          let amount = creep.room.getHarvesterAmount();
+          amount += Math.floor(resourceAtPosition / config.carry.callHarvesterPerResources);
```
Nevrdid: better Math.ceil i think

src/prototype_creep_resources.js
kahojyun: Carries and harvesters will only transfer to structures in the base, so it's unnecessary to check `STRUCTURE_CONTAINER` now.  We have `terminalAvailable` for `STRUCTURE_TERMINAL`.

src/role_carry.js
kahojyun: The check is included in `transferToStructures`.  If `transferred` is `false`, carry can't transfer to any structures including storage.

src/role_carry.js
kahojyun: yeah, but this is irrelevant to my pr...  We can start a new pr to fix this and `getHarvesterAmount` in `prototype_room_my.js`.

src/role_carry.js
```diff
@@ -143,18 +143,20 @@ roles.carry.preMove = function(creep, directions) {
...
+      } else if (creep.memory.routing.pathPos === 0) {
         creep.drop(RESOURCE_ENERGY);
         reverse = false;
-
-        let resourceAtPosition = 0;
-        const resources = creep.pos.lookFor(LOOK_RESOURCES);
-        for (const resource of resources) {
-          resourceAtPosition += resource.amount;
+        const storage = creep.room.storage;
+        if (!(storage && storage.my && storage.isActive())) {
```
mschultheiss83: storage.isActive() has medium cost, we should cache for some ticks

democratic-collaboration:
A new review, yeah.

    Votes: 208/504
    Coefficient: 0.412698412698
    Merging in 4 days 0 hours
    Age 1 days 20 hours

src/role_carry.js
Nevrdid: alright, missed it

democratic-collaboration:
A new review, yeah.

    Votes: 333/505
    Coefficient: 0.659405940594
    Merging in 1 days 10 hours
    Age 1 days 22 hours

src/prototype_creep_resources.js
```diff
@@ -360,7 +360,9 @@ const filterTransferrables = function(creep, object) {
...
+    _.sum(object.store) + _.sum(creep.carry) > object.storeCapacity :
+    object.energy === object.energyCapacity) {
```
sparr: This whole energy-vs-store needs to be moved out to a helper function. It's used multiple places in the code, and ripe for errors.

src/role_carry.js
kahojyun: I don't think it a problem because `isActive` is called only when a carry stands on `pathStart`.

PR #414, opened 2017-09-17
Memory segments

samogot:
Memory segments usage implementation 
Use segments for cost matrices


samogot:
@TooAngel @Nevrdid Reviews appreciated

PR #422, opened 2017-09-22
added config.room.observerRange

mschultheiss83:
just added config.room.observerRange (1 to 10) to reduce memory

### DIFF
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -128,7 +128,7 @@ Room.prototype.handleObserver = function() {
 
     // TODO scan full range, first implementation
     const nameSplit = this.splitRoomName();
-    const observerRange = (config.room.observerRange <= 10 && config.room.observerRange > 0) ? config.room.observerRange : OBSERVER_RANGE;
+    const observerRange = Math.max(Math.min(config.room.observerRange, OBSERVER_RANGE), 1);
     const fullLength = 2 * observerRange + 1;
     const numberOfFields = fullLength * fullLength;
     const offset = Game.time % numberOfFields;
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -128,13 +128,14 @@ Room.prototype.handleObserver = function() {
 
     // TODO scan full range, first implementation
     const nameSplit = this.splitRoomName();
-    const fullLength = 2 * OBSERVER_RANGE + 1;
+    const observerRange = (config.room.observerRange <= 10 && config.room.observerRange > 0) ? config.room.observerRange : OBSERVER_RANGE;
+    const fullLength = 2 * observerRange + 1;
     const numberOfFields = fullLength * fullLength;
     const offset = Game.time % numberOfFields;
-    const xOffset = Math.floor(offset / fullLength) - OBSERVER_RANGE;
-    const yOffset = Math.floor(offset % fullLength) - OBSERVER_RANGE;
-    let xPos = +nameSplit[2] + xOffset;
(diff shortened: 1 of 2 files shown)


src/prototype_room_my.js
```diff
@@ -128,13 +128,14 @@ Room.prototype.handleObserver = function() {
 
     // TODO scan full range, first implementation
     const nameSplit = this.splitRoomName();
-    const fullLength = 2 * OBSERVER_RANGE + 1;
+    const observerRange = (config.room.observerRange <= 10 && config.room.observerRange > 0) ? config.room.observerRange : OBSERVER_RANGE;
```
sparr: `config.room.observerRange <= OBSERVER_RANGE` instead? in case it changes, or is different on a private server. also nice even if only to remove magic numbers.

sparr: also, if config.room.observerRange somehow gets set to -1, do we really want to bump it to 10 instead of to 0?

TooAngel: `const observerRange = Math.min(config.room.observerRange, OBSERVER_RANGE)`

sparr: just min() ignores the possibility that config.room.observerRange is negative. should we just say that's an error condition, or should we handle that gracefully here if something else has programmatically set that config lower than 0?

Nevrdid: why not : 
`config.room.observerRange === -1 ? OBSERVER_RANGE : Math.min(config.room.observerRange, OBSERVER_RANGE);`

sparr: `config.room.observerRange < 0 ? OBSERVER_RANGE : Math.min(config.room.observerRange, OBSERVER_RANGE);`

Nevrdid: why set it up to a value negative if we want 0 ? just setup directly 0 ^^ we could be able to set it up to -1 for use constant one

sparr: I am predicting a scenario where someone uses code to calculate their desired config.room.observerRange dynamically, and that code unexpectedly sets it to an arbitrary negative number.

I would rather use `observerRange: OBSERVER_RANGE` in a config if I want the constant one.

Nevrdid: ok but shouldn't be difficult to even cap the result of any getObserverRange method. agree for use constant in config directly ^^. then just a Math.min should be enouth

TooAngel: I wouldn't do any magic. If `< 0` set to `OBSERVER_RANGE`.

democratic-collaboration:
A new review, yeah.

    Votes: 329/505
    Coefficient: 0.651485148515
    Merging in 6 days 13 hours
    Age 0 days 9 hours

src/prototype_room_my.js
```diff
@@ -128,13 +128,14 @@ Room.prototype.handleObserver = function() {
 
     // TODO scan full range, first implementation
     const nameSplit = this.splitRoomName();
-    const fullLength = 2 * OBSERVER_RANGE + 1;
+    const observerRange = Math.max(Math.min(config.room.observerRange, OBSERVER_RANGE), 1);
```
sparr: Why 1 instead of 0? Is there another way to disable observer scanning, or would a user want to set config.room.observerRange to zero to disable their scanning?

TooAngel: True, this should be `0`

Reducing to 1 commit is appreciated, too

democratic-collaboration:
A new review, yeah.

    Votes: 396/513
    Coefficient: 0.771929824561
    Merging in 1 days 2 hours
    Age 3 days 11 hours

PR #432, opened 2017-09-23
move pathStart from first pos near storage to best

sparr:
pathStart was previously placed at the first available (in direction order) location adjacent to the storage. let's switch that to the location with the most adjacent open spaces to avoid some traffic jams.

### DIFF
--- a/src/prototype_room_init.js
+++ b/src/prototype_room_init.js
@@ -46,7 +46,7 @@ Room.prototype.initSetStorageAndPathStart = function() {
   costMatrix.set(storagePos.x, storagePos.y, config.layout.structureAvoid);
   this.setMemoryCostMatrix(costMatrix);
 
-  this.memory.position.creep.pathStart = storagePos.getFirstNearPosition();
+  this.memory.position.creep.pathStart = storagePos.getBestNearPosition();
 
   const route = [{
     room: this.name,


sparr:
@samogot the traffic jam this aims to address is the one where pathstart is stuck in a corner so every path comes from there, leading to a lot of overlapping/intersecting paths that don't need to be.

Here's a layout with pathstart north of storage which leads to every path overlapping at the spot northwest of storage. This patch moves that pathstart to southwest of storage.

samogot:
I'd never choose such room)) It's obvious that source near to controller will cause problems and needs some special handling.

sparr:
@samogot special handling like #435 ? :)

TooAngel:
Regarding the room layout versioning. https://github.com/TooAngel/screeps/blob/2dddbb4f028372163476a66f013fc50d1183d0ed/src/config.js#L204

There is already a versioning in place, but as soon as the room layout is renewed the new version will be used, force a rebuild by increasing the version

sparr:
I'm thinking of versioning each room, and not updating rooms by default, unless a config option is set to update all rooms, or a single room has its version manually bumped.

TooAngel:
The current logic is:
When the layout is generated the current version is stored.
If the room version and the config version differs, the layout is redone.
If we make sure the room layout is only built the first time the room is used (I'm currently not so sure this is the case).

I think this should be similar to your idea?

sparr:
Marking this WIP. Will get combined into a single new-base-layout patch to avoid rebuilding existing bases.

PR #435, opened 2017-09-23
[WIP] sourcer near storage deposits directly, no carry

sparr:
The goal here is for a sourcer adjacent to the storage to deposit energy directly to the storage, without needing a carry. This is especially important if the carry would insist on standing in a spot that causes a traffic jam.

[WIP] pending more testing and possibly more improvements.

### DIFF
--- a/src/prototype_creep_harvest.js
+++ b/src/prototype_creep_harvest.js
@@ -14,8 +14,17 @@ Creep.prototype.handleSourcer = function() {
 
   this.buildContainer();
 
-  if (!this.room.controller || !this.room.controller.my || this.room.controller.level >= 2) {
-    this.spawnCarry();
+  const nearStorage = this.room.storage && this.pos.isNearTo(this.room.storage);
+
+  if (nearStorage) {
+    const workParts = this.body.filter((part) => part.type === WORK).length;
+    if (_.sum(this.carry) > this.carryCapacity - workParts * 2) {
+      this.transfer(this.room.storage, RESOURCE_ENERGY);
+    }
+  } else {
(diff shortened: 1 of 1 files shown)


TooAngel:
The layout is currently. Each `sourcer` gets a `link`, which transfers the energy to the link at the storage. As long as there is no link, the `sourcer` simply drops the energy and the call carry mechanism drops in.

Pretty basic fits for most of the rooms. What you are describing is a special case within this logic,

We had similar cases where carries could block other creeps. I think a better approach if the carry wouldn't stand still, but flips between two positions while waiting on energy. This would allow other creeps to pass by.

sparr:
So far this patch just has the sourcer deposit directly to the storage and not summon/spawn a carry. Thanks for the heads up that I need to have the layout skip one of the links and the sourcer not expect it, too. I'll have to wait until I have a room with links to test that.

src/prototype_creep_harvest.js
```diff
@@ -14,8 +14,17 @@ Creep.prototype.handleSourcer = function() {
 
   this.buildContainer();
 
-  if (!this.room.controller || !this.room.controller.my || this.room.controller.level >= 2) {
-    this.spawnCarry();
+  const nearStorage = this.room.storage && this.pos.isNearTo(this.room.storage);
+
+  if (nearStorage) {
+    const workParts = this.body.filter((part) => part.type === WORK).length;
+    if (_.sum(this.carry) > this.carryCapacity - workParts * 2) {
```
mschultheiss83: why not 
```
if (_.sum(this.carry) > 0) {
```

PR #440, opened 2017-09-23
carry with misplacedSpawn will fill storage in addition to extensions

sparr:


src/role_carry.js
```diff
@@ -69,7 +69,7 @@ roles.carry.handleMisplacedSpawn = function(creep) {
     //     creep.say('cmis', true);
     if (creep.carry.energy > 0) {
       const structure = creep.pos.findClosestByRange(FIND_MY_STRUCTURES, {
-        filter: (object) => object.energy < object.energyCapacity,
+        filter: (object) => ((object.store ? _.sum(object.store) < object.storeCapacity : object.energy < object.energyCapacity)),
```
sparr: Is there a scenario where an owned room will have containers at the sourcer that I need to exclude here?

TooAngel: We do not build containers at the sourcer in our own rooms.

The only case I could come up is your first room, where you switched from another AI, or if in a room we claim already container exist. I don't think this is necessary - would be better, but not necessary.

democratic-collaboration:
A new review, yeah.

    Votes: 398/515
    Coefficient: 0.772815533981
    Merging in -1 days 8 hours
    Age 2 days 21 hours

PR #441, opened 2017-09-23
carry in misplacedSpawn picks up dropped energy at sourcer then bails

sparr:
Previously a carry operating in a misplacedSpawn room would seek out the sourcer then let all its other role code handle picking up energy, path reversal, etc. This code replaces that with a much more straightforward "pick up dropped energy, then bail out". May need to be expanded on if a link or container could still be in play at this point?

### DIFF
--- a/src/role_carry.js
+++ b/src/role_carry.js
@@ -88,9 +88,13 @@ roles.carry.handleMisplacedSpawn = function(creep) {
         creep.moveTo(sourcePos, {
           ignoreCreeps: true,
         });
-        if (creep.pos.getRangeTo(sourcePos) > 1) {
-          return true;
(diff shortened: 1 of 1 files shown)


Nevrdid:
if i understand it right, you are just skipping all carrys action when carry is going to source.

sparr:
Yes. This was per @TooAngel saying that the pickup could happen in the misplaced spawn handler instead of. Something about a room with a misplaced spawn was causing my carries to stop processing in the post-action handler steps, leading to them not picking up.

sparr:
The misbehavior here has been mostly fixed by other recent merges. I am closing this PR. We should probably still simplify the handling of this scenario, but this PR isn't good enough.

PR #443, opened 2017-09-24
only update memory.last when fatigue === 0

sparr:
A creep with low MOVE or walking through swamp isn't stuck just because it's been sitting still for three ticks. It's stuck if it's sitting still for three ticks after its fatigue reaches zero.

### DIFF
--- a/src/prototype_creep.js
+++ b/src/prototype_creep.js
@@ -133,15 +133,17 @@ Creep.prototype.handle = function() {
     this.log(message);
     Game.notify(message, 30);
   } finally {
-    if (this.memory.last === undefined) {
-      this.memory.last = {};
+    if (this.fatigue === 0) {
+      if (this.memory.last === undefined) {
+        this.memory.last = {};
+      }
+      const last = this.memory.last;
+      this.memory.last = {
+        pos1: this.pos,
+        pos2: last.pos1,
+        pos3: last.pos2,
+      };
     }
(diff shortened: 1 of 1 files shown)


Nevrdid:
desapprove as it will no more work when a creep is stuck ( pos == pos1 != pos2 != pos3) and will for sure break nearly all use of isStuck(). Creeps shouldn't need more than 3 ticks for move or we should use another system.
Here is how i did for my scouts : https://github.com/Nevrdid/screeps/blob/Workbranch2/src/role_scout.js#L198

sparr:
@Nevrdid if a creep makes one step onto swamp, it will not be able to move for 5-10(?) ticks. It's not stuck, that's just how slow walking through swamp is. I don't want isStuck() to return true for a creep that has been sitting on a swamp tile waiting for its fatigue counter to reach zero.

Also a creep with low MOVE, like a sourcer or upgrader, might move less than one tile every 3 ticks. Again, not stuck, just moving slow by design.

Nevrdid:
yeah, may be true and surely better to count with fatigue but as you did, the last positions will always be late setup as creep will have to wait to fatigue come back to zero for add last pos. I think it's better always update it maybe when fatigue rise back or using a flag in memory as i did.

Nevrdid:
as for exemple, still upgrade last each tick no care about fatigue and return true on isStuck() only if fatigue was null at least once since it's stuck

PR #446, opened 2017-09-24
carry stops before standing on sourcer position

sparr:
Previously, if a carry was going to get energy from a sourcer, but the sourcer was AWOL, then the carry would stand in the sourcer's spot and block the sourcer from getting to work.


Nevrdid:
x/ i supose you using https://github.com/TooAngel/screeps/pull/441/files which is probably the cause.
The issue is that actually and it work well. when a carry reach endOfPath, it pass to action which made it reverse. But with your other PR, it broke that.

sparr:
Closing this for now. Other recent updates may have resolved these issues.

PR #447, opened 2017-09-24
don't skip ticks if tickLimit is unusually high (private servers?)

sparr:
ScreepsPl.us has a Game.cpu.tickLimit of 10100 for some weird reason

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -29,7 +29,7 @@ if (config.profiler.enabled) {
 }
 
 const main = function() {
-  if (Game.cpu.bucket < 2 * Game.cpu.tickLimit) {
+  if (Game.cpu.bucket < 2 * Game.cpu.tickLimit && Game.cpu.bucket < Game.cpu.limit * 10) {
     console.log('Skipping tick ' + Game.time + ' due to lack of CPU.');
     return;
   }


src/main.js
```diff
@@ -29,7 +29,7 @@ if (config.profiler.enabled) {
 }
 
 const main = function() {
-  if (Game.cpu.bucket < 2 * Game.cpu.tickLimit) {
+  if (Game.cpu.bucket < 2 * Game.cpu.tickLimit && Game.cpu.tickLimit < 3000) {
```
Nevrdid: not sure it's good to allow all executions : 
i made it like this personally.
```
if (Game.cpu.bucket < 2 * (Game.cpu.tickLimit < 1000 ? Game.cpu.tickLimit : 1000)) {
    console.log('Skipping tick ' + Game.time + ' due to lack of CPU. Limit:' + Game.cpu.tickLimit + ' - Bucket: ' + Game.cpu.bucket);
    return;
  }
```

TooAngel: Would like to open the discussion:
Shouldn't we just use `Game.cpu.limit` (`2 *`). The AI should stay under the CPU limit, the difference to `tickLimit` can be a buffer for tasks which are executed only rarely.

samogot: I think we shouldn't. I've faced cpu problems a lot, there was tick that costs more than `Game.cpu.limit * 2`  in total. Especially ticks after clearing room memory. Periods of room layout debugging was very heavy, and before `RoomPosition.inPath` (#384) improvement, even full `tickLimit` wasn't enough to compute layout for some rooms. Also in worst times before memory segments (#414) there was situations where only one tick of five wasn't skipped. Poth problems are fixed (but not merged yet), but if we'd face something like this in future, `Game.cpu.limit` for sure wound't be enough

samogot: from the other hand, I see no reason to have forcibly `tickLimit * 2` in bucket, as only `tickLimit` may be used in one tick

Nevrdid: yeah, it can be improve for sure, I supose we just have to define some task priority and then use separate checks for skip them. However, on my branch, i already adapted amount of reserved / new rooms to bucket and it immediatly solved all bucket issues i had.. Also, scout is actually a big cpu eater.

TooAngel: It is all about how much is in the bucket and my suggestion would be `Game.cpu.bucket < 2 * Game.cpu.limit`. Which prevents any issues with `tickLimit` unusually high and we just say double the amount we can savely use.
The are some rare cases where we would skip even if the `tickLimit` wouldn't be reached, IMHO acceptable.

Nevrdid: I don't think there is actually big cpu burn problem, skip happen only when codebase tried to manage too much things at same time for a long period. Rooms layout and others few ticks operations never emptied my bucket.

samogot: I cant agree with you. I'm facing cpu problems from time to time since I started using TooAngel not GCL 2. I'm now on GCL 7 and I always have maximum amount of rooms, and at least 3 remote sourcers. Maybe everything is ok on high GCL, but I had a lot of problems, and tried to solve them)

samogot: I consider the feature of having at least `tickLimit` cpu in bucket as reasonable. In case of emergency, if something have fucked up, it's better to skip some ticks and have ability to perform as long computation as possible, instead of trying to do it on every tick, get execution interruption and either never get out of this loop, or execute something partially and have chance to get memory corruption.

samogot: Also, I agree that it would be better to have feature priority and skip part of tick instead of whole tick

samogot: But from the other hand - we shouldn't skip anything at all. And when we skip whole tick - problem is more obvious

TooAngel: Agree we shouldn't need that check, but we are currently close (or over) the cpu limit.

Would like to steer the discussion back to the current issue, to solve this soonish:

As it sounds the current logic is bad if the server has `Game.cpu.tickLimit` of 10100 (high value).
We skip ticks by ourself to not run into the server side execution blocking due to moverly exceeding the CPU time.
Properly handling the CPU / Bucket logic ich much more complex also regarding the frequency of specific bursts.

If I understood it correctly we can use up to `Game.cpu.ticklimit - Game.cpu.limit` from the bucket in each tick.
This wouldn't solve the current issue. Taking any random value would run into problems with different private server configurations. The max ticklimit number is not available as constant.

Tbh I still have a good solution. I'm still in favor of `2 * Game.cpu.limit`, we shouldn't use more than the double of our cpu limit. I know it is not the best argument. This solves the problem and I think better than `2 * Game.cpu.ticklimit`.

sparr:
Updated this PR. Now instead of checking against an arbitrary bucket capacity I'm checking against 10*Game.cpu.limit. So, if there's a private server that actually allows using more than 9 ticks worth of cpu out of the bucket at once, we will just decline to do so.

democratic-collaboration:
A new review, yeah.

            Votes: -53/530
            Coefficient: -0.1
            Merging in 10 days 22 hours
            Age 0 days 1 hours

democratic-collaboration:
A new review, yeah.

            Votes: 187/530
            Coefficient: 0.352830188679
            Merging in 6 days 8 hours
            Age 0 days 2 hours

democratic-collaboration:
A new review, yeah.

            Votes: 196/531
            Coefficient: 0.369114877589
            Merging in 4 days 9 hours
            Age 1 days 22 hours

PR #394, opened 2017-09-09
Refactor move by path my to avoid duplication and cycles in profiler graph

samogot:

PR #448, opened 2017-09-25
fix followPath

Nevrdid:
The PR should fix alot of recent seen strange behaviors. Please merge it and look if your fixes are still needed.

### DIFF
--- a/src/prototype_creep_routing.js
+++ b/src/prototype_creep_routing.js
@@ -162,14 +162,25 @@ Creep.prototype.getDirections = function(path, pathPos) {
   };
 };
 
+Creep.prototype.checkPreMove = function(datas) {
+  if (!datas.path) {
+    return true;
+  }
+  if (datas.pathPos < 0) {
+    this.memory.routing.pathPos = datas.pathPos;
+    return true;
+  }
+  return false;
+};
+
 Creep.prototype.followPath = function(action) {
   const route = this.getRoute();
   const routePos = this.getRoutePos(route);
 
   // TODO Disable base room for now
   // if (routePos === 0) {
-  // this.say('R:Base');
-  // return false;
+  //   this.say('R:Base');
+  //   return false;
   // }
 
   if (!this.memory.routing.targetId && this.room.name === this.memory.routing.targetRoom) {
@@ -177,10 +188,17 @@ Creep.prototype.followPath = function(action) {
     return action(this);
   }
   const prepareData = this.moveByPathPrepare(route, routePos, 'pathStart', this.memory.routing.targetId);
-  if (prepareData.unit.preMove) {
-    if (prepareData.unit.preMove(this, prepareData.directions)) {
+  if (this.checkPreMove(prepareData)) {
+    if (prepareData.unit.preMove && prepareData.unit.preMove(this, prepareData.directions)) {
       return true;
     }
+    if (!prepareData.path || prepareData.path.length === 0) {
+      return false;
+    }
+  } else if (prepareData.unit.preMove && prepareData.unit.preMove(this, prepareData.directions)) {
+    // if(!this.isStuck()){  // May be used if a issue happen. Didn't need it while testing
+    return true;
+    // }
   }
   return this.moveByPathMy(route, routePos, 'pathStart', this.memory.routing.targetId, action, prepareData);
 };
@@ -188,6 +206,7 @@ Creep.prototype.followPath = function(action) {
 Creep.prototype.moveByPathPrepare = function(route, routePos, start, target) {
   const result = {};
   result.unit = roles[this.memory.role];
+
   // Somehow reset the pathPos if the path has changed?!
   result.path = this.room.getPath(route, routePos, start, target);
   if (!result.path) {
@@ -206,26 +225,8 @@ Creep.prototype.moveByPathMy = function(route, routePos, start, target, action, 
     prepareData = this.moveByPathPrepare(route, routePos, start, target);
   }
   const {unit, path, pathPos, directions} = prepareData;
-  if (!path) {
-    // TODO this could be because the targetId Object does not exist anymore
-    // this.log('newmove: no path legacy fallback: ' + this.memory.base + ' ' +
-    // this.room.name + ' ' + this.memory.base + ' ' +
-    // this.memory.routing.targetRoom + ' routePos: ' + routePos + ' route: ' +
-    // JSON.stringify(route));
-    this.say('R:no path');
-    this.log('R:no path');
-    // this.log('R:no path: pathStart-' + this.memory.routing.targetId);
-    return false;
-  }
 
   if (pathPos < 0) {
-    // this.say('R:pos -1');
-    this.memory.routing.pathPos = pathPos;
-    if (path.length === 0) {
-      this.log('config_creep_routing.followPath no pos: ' + JSON.stringify(path));
-      return false;
-    }
-
     let posFirst;
     try {
       if (this.memory.routing.reverse) {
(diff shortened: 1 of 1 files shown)


src/prototype_creep_routing.js
```diff
@@ -177,27 +177,31 @@ Creep.prototype.followPath = function(action) {
...
+  if (result.path) {
+    result.pathPos = this.getPathPos(route, routePos, result.path);
+    if (result.pathPos > -1) {
+      result.directions = this.getDirections(result.path, result.pathPos);
+    }
   }
-  result.pathPos = this.getPathPos(route, routePos, result.path);
-  if (result.pathPos < 0) {
-    return result;
+  if ((result.unit.preMove && result.unit.preMove(this, result.directions)) || !result.path) {
```
samogot: we mustn't call preMove here, because this may potentially lead to infinite recursion:
harvester.preMove->...->moveByPathMy->moveByPathPrepare->harvester.preMove
this was the reason of extracting moveByPathPrepare

democratic-collaboration:
A new review, yeah.

    Votes: 1/507
    Coefficient: 0.00197238658777
    Merging in 9 days 19 hours
    Age 0 days 4 hours

src/prototype_creep_routing.js
Nevrdid: preMove shouldn't call moveByPathMY
if a role do it then it is that which have to be fixed

samogot: even if it never goes into recursion now because of some conditions, we should only call preMove from followPath, to have neat call  graph in recursion, and to not fail into recursion if we'd accidentally break conditions that prevent it

samogot: @Nevrdid, maybe, but it is called now. there was `noPreMove` flag, before `moveByPathPrepare` extraction

Nevrdid: ok will look at how to export it

samogot: #394

Nevrdid: nopremove flag ?

Nevrdid: yeah seen it else i shouldn't say it come from here :)  don't you see these preMove you just removed?

samogot: yeah, that was `skipPreMove`
I suppose that I've failed and introduce regression while refactoring, but I still don't see where is it)

Nevrdid: alright, exported checks in a new prototype. do you seen betetr the difference ?

samogot: nope. 😞 
I've figured out that I had messed up with order of calling `preMove` and setting `memory.routing.pathPos`, but it not looks like this it the thing you've fixed. All other conditional branches, as far as I can see will lead to the same result. I mean that they shouldn't be the same, as you've fixed something, but I can't figure out what it was)
Maybe you'd just describe what branch if failed in my change?)

samogot: Ah! I found something... After you change we return `true` instead of `false` in case of `datas.pathPos < 0 && datas.path.length === 0`. But we haven't this behavior before my change)

Nevrdid: yeah ^^' sorry didn't respond as i fixing it again ^^' as simplifying it i broke it again x/

Nevrdid: the problem is followPath returns which differ depending prepare datas. but we can't really do a returnFalse() method as we have to altern differents checks.  will push back soon.

Nevrdid: alright ^^ fixed.

samogot: makes more sense now. so setting `memory.routing.pathPos` before `preMove` was really the case.

src/prototype_creep_routing.js
```diff
@@ -162,42 +162,62 @@ Creep.prototype.getDirections = function(path, pathPos) {
...
+
   // Somehow reset the pathPos if the path has changed?!
   result.path = this.room.getPath(route, routePos, start, target);
-  if (!result.path) {
-    return result;
-  }
-  result.pathPos = this.getPathPos(route, routePos, result.path);
-  if (result.pathPos < 0) {
-    return result;
+  if (result.path) {
```
TooAngel: nitpicking: I like the previous format more.

src/prototype_creep_routing.js
```diff
@@ -162,42 +162,62 @@ Creep.prototype.getDirections = function(path, pathPos) {
...
     return action(this);
   }
   const prepareData = this.moveByPathPrepare(route, routePos, 'pathStart', this.memory.routing.targetId);
-  if (prepareData.unit.preMove) {
-    if (prepareData.unit.preMove(this, prepareData.directions)) {
+  if (this.checkPreMove(prepareData)) {
+    if (prepareData.unit.preMove && prepareData.unit.preMove(this, prepareData.directions)) {
       return true;
     }
+    if (!this.checkContinue(prepareData)) {
```
samogot: Do we really need to extract `checkContinue`? I think it would be easier to understand it inline

src/prototype_creep_routing.js
```diff
@@ -206,26 +226,9 @@ Creep.prototype.moveByPathMy = function(route, routePos, start, target, action,
...
-    // JSON.stringify(route));
-    this.say('R:no path');
-    this.log('R:no path');
-    // this.log('R:no path: pathStart-' + this.memory.routing.targetId);
-    return false;
-  }
 
   if (pathPos < 0) {
-    // this.say('R:pos -1');
     this.memory.routing.pathPos = pathPos;
```
samogot: do we still need this line here?

democratic-collaboration:
A new review, yeah.

    Votes: 127/509
    Coefficient: 0.249508840864
    Merging in 11 days 5 hours
    Age 0 days 0 hours

src/prototype_creep_routing.js
Nevrdid: yeah, will restore it. switched when i included checks into prepare method as i couldn't return in middle of method.

src/prototype_creep_routing.js
Nevrdid: 👍

Nevrdid:
I believe it's fixed back ^^' reviews and test welcomed.

src/prototype_creep_routing.js
```diff
@@ -162,32 +162,51 @@ Creep.prototype.getDirections = function(path, pathPos) {
...
-  if (prepareData.unit.preMove) {
-    if (prepareData.unit.preMove(this, prepareData.directions)) {
+  if (this.checkPreMove(prepareData)) {
+    if (prepareData.unit.preMove && prepareData.unit.preMove(this, prepareData.directions)) {
       return true;
     }
+    if (!prepareData.path || prepareData.path.length === 0) {
+      return false;
+    }
+  } else if (prepareData.unit.preMove && prepareData.unit.preMove(this, prepareData.directions)) {
```
samogot: hmmm... I do not like this code duplication. Looking at it now, do we really need `checkPreMove`? We should run `preMove` both in case of `true` and `false`. Maybe it would be better to inline it? Or even move `this.memory.routing.pathPos = datas.pathPos` line into `moveByPathPrepare` function?

samogot: I'm sorry I'm too annoying, even when it's my regression is being fixed)

democratic-collaboration:
A new review, yeah.

    Votes: 130/516
    Coefficient: 0.251937984496
    Merging in 6 days 7 hours
    Age 1 days 4 hours

democratic-collaboration:
A new review, yeah.

            Votes: 130/527
            Coefficient: 0.246679316888
            Merging in 4 days 4 hours
            Age 3 days 7 hours

democratic-collaboration:
A new review, yeah.

            Votes: 161/529
            Coefficient: 0.304347826087
            Merging in 3 days 1 hours
            Age 3 days 20 hours

PR #451, opened 2017-09-28
new tower placement algorithm, fixes bugs and improves

### LINKED ISSUE #445: tower link placement mistakes
https://i.imgur.com/6qfdLZZ.png
https://i.imgur.com/xYLx3og.png

two example so far. I'll try to debug the code when I get a base far enough to build links.

sparr:
This PR introduces a new helper function, Room.getMemoryPathSidewalk(). This produces a path-like list that contains every position on one side of a memorized path. This function is then used to fix the tower/filler/link placement code to avoid the previous problems where there would be gaps between those positions. I also included improvements like using both sides of the road and packing the positions closer together on diagonal roads.

Example of new layout:

Fixes #445

### DIFF
--- a/src/prototype_room_memory.js
+++ b/src/prototype_room_memory.js
@@ -152,6 +152,93 @@ Room.prototype.getMemoryPath = function(name) {
 };
 
 /**
+ * Returns a pseudo-path including every tile adjacent to one side of the given path
+ *
+ * @param {String} name - the name of the path
+ * @param {Number} side - +1 or -1 indicating which side of the path
+ * @return {array|boolean} path
+ */
+Room.prototype.getMemoryPathSidewalk = function(name, side = 1) {
+  const path = this.getMemoryPath(name);
+  if (!path) {
+    return false;
+  }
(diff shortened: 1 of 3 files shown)


src/prototype_room_memory.js
```diff
@@ -152,6 +152,93 @@ Room.prototype.getMemoryPath = function(name) {
 };
 
 /**
+ * Returns a pseudo-path including every tile adjacent to one side of the given path
```
samogot: this function should be placed in _routing or somewhere else, but not in _memory file

sparr:
Marking this WIP. Will get combined into a single new-base-layout patch to avoid rebuilding existing bases.

src/prototype_room_memory.js
sparr: I'm conflicted. It's theoretically a routing function, but the only place it's currently called is in_memory. I decided to put it here until/unless it gets used somewhere else. I could easily be convinced otherwise.

PR #452, opened 2017-09-28
add options to validPosition(), ignore paths when placing sourcers

sparr:
Currently attempting to re-place creep positions in a room is not idempotent because they will avoid the paths previously planned to reach them. This patch introduces a way to pass options down through the relevant calls so that validPosition can be told to ignore specific position features.


samogot:
why do you need to re-place only creep positions without everything else? `Room.prototype.clearMemory()` works perfectly well.

sparr:
`clearMemory()` has many side effects.

PR #455, opened 2017-09-29
catch exceptions in the visualizer

sparr:
There's other code around this that I want to rearrange, maybe optimize, but just having exceptions here be caught is a nice upgrade for future development/debugging efforts.

Maybe we should be printing out the call trace like the normal exception handler?

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -41,18 +41,27 @@ const main = function() {
     brain.handleIncomingTransactions();
     brain.handleQuests();
   } catch (e) {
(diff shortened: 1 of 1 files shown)


TooAngel:
I like, stacktraces would be good

democratic-collaboration:
A new review, yeah.

            Votes: 12/531
            Coefficient: 0.0225988700565
            Merging in 7 days 20 hours
            Age 1 days 21 hours

PR #472, opened 2017-10-23
Restrict new room claiming on `cpu.limit`

TooAngel:
Introduce `config.nextRoom.cpuPerRoom` to be able to set min CPU
per room and avoid claiming of new rooms.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -90,6 +90,7 @@ global.config = {
     numberOfNextroomers: 10,
     nextroomerInterval: 500,
     maxRooms: 20,
+    cpuPerRoom: 13, // Necessary CPU per room, prevent claiming new rooms
     revive: true,
     maxDistance: 17,
     minNewRoomDistance: 2,
--- a/src/config_brain_nextroom.js
+++ b/src/config_brain_nextroom.js
@@ -1,7 +1,10 @@
 'use strict';
 
 brain.handleNextroom = function() {
-  if (Memory.myRooms && Memory.myRooms.length < Game.gcl.level && Memory.myRooms.length < config.nextRoom.maxRooms) {
+  if (Memory.myRooms &&
+    Memory.myRooms.length < Game.gcl.level &&
+    Memory.myRooms.length < config.nextRoom.maxRooms &&
+    (Memory.myRooms.length + 1) * config.nextRoom.cpuPerRoom < Game.cpu.limit) {
     if (Game.time % config.nextRoom.ttlPerRoomForScout === 0) {
       for (const roomName of Memory.myRooms) {
         const room = Game.rooms[roomName];


src/config.js
```diff
@@ -90,6 +90,7 @@ global.config = {
     numberOfNextroomers: 10,
     nextroomerInterval: 500,
     maxRooms: 20,
+    cpuPerRoom: 12, // Necessary CPU per room, prevent claiming new rooms
```
mschultheiss83: i just did a respawn on monday, on my first room i had cpu peeks up to 55 cpu, with 2 i have an average so this could be too low? 
// so is cpuPerRoom: 20 too high? 
// or so cpuPerRoom: _.floor(Game.cpu.limit - 20) / 5,

@TooAngel just btw should not be maxRooms: Game.gcl.level, ?

TooAngel: `maxRooms` was my first approach to reduce the number of rooms to fit somehow to my `cpu.limit` level and to introduce some mechanism for skipping.
`cpuPerRoom` is my next approach, next step, measure the CPU usage and only claim a room when enough CPU is available.

I'm thinking about increasing to 13. You start with 30 CPU and get per each GCL one more.
With 20 you can have 3 rooms and on GCL 4 you have to wait a full gcl, IMHO this is too early, especially for the bot or new players.
I think ~13 CPU is a limit we could eye for in regards to CPU optimizations. What do you think?

democratic-collaboration:
A new review, yeah.

            Votes: 277/548
            Coefficient: 0.505474452555
            Merging in 3 days 0 hours
            Age 1 days 22 hours

src/config_brain_nextroom.js
```diff
@@ -1,7 +1,10 @@
 'use strict';
 
 brain.handleNextroom = function() {
-  if (Memory.myRooms && Memory.myRooms.length < Game.gcl.level && Memory.myRooms.length < config.nextRoom.maxRooms) {
+  if (Memory.myRooms &&
+    Memory.myRooms.length < Game.gcl.level &&
+    Memory.myRooms.length < config.nextRoom.maxRooms &&
+    Memory.myRooms.length * config.nextRoom.cpuPerRoom < Game.cpu.limit) {
```
Evanito: Should be
`(Memory.myRooms.length + 1) * config...`
Otherwise it will go over the limit when colonizing the new room.

Otherwise having a single "unallocated" CPU left will allow colonization of a new room, and starve all rooms as a result.

democratic-collaboration:
A new review, yeah.

            Votes: 278/549
            Coefficient: 0.506375227687
            Merging in 1 days 20 hours
            Age 3 days 2 hours

Evanito:
Can the Democratic Collab bot not tell the difference between an approve and request changes?

src/config_brain_nextroom.js
TooAngel: Not sure if I can follow.

As soon as a controller is claimed the room will end up in `myRooms`. There are some reasons how it can happen that it will claim more rooms, if you GCL is higher, but I don't see the point in that change.

TooAngel:
> Edit: Looking at the code, it should be able to tell the difference. Is the bot not running the same version?

Need to check that, maybe it is because you approved first, and requested changes later.

TooAngel:
On https://dc.tooangel.de/TooAngel/screeps/pull/472 it looks fine.

src/config_brain_nextroom.js
Evanito: The idea of this change is to reserve at least 13 (by default) cpu per room so that each room has enough CPU to run properly, right?

What I'm noticing is that the code doesn't take the new room into account when doing this calculation (before claiming.)

If I theoretically have 14 cpu and one room, this will allow a new room to be claimed even though there would not be sufficient CPU to maintain both rooms, which I feel is unexpected behavior.

TooAngel: So I have one room: 1 * 13 = 13 Cpu limit 30

TooAngel: you are right

democratic-collaboration:
A new review, yeah.

            Votes: 280/551
            Coefficient: 0.508166969147
            Merging in 4 days 12 hours
            Age 0 days 9 hours

democratic-collaboration:
A new review, yeah.

            Votes: 252/551
            Coefficient: 0.457350272232
            Merging in 4 days 7 hours
            Age 1 days 2 hours

PR #474, opened 2017-10-23
Introduce creepLog, config based logging of creeps

TooAngel:
`config.creepLog.roles` and `config.creepLog.rooms` to filter specific
roles and rooms.

### DIFF
--- a/src/logging.js
+++ b/src/logging.js
@@ -5,3 +5,21 @@ Room.prototype.log = function(...messages) {
 RoomObject.prototype.log = function(...messages) {
   console.log(`${Game.time} ${this.room.name.rpad(' ', 6)} ${this.name.rpad(' ', 20)} ${this.pos} ${messages.join(' ')}`);
 };
+
+/*
+ * Log creep message based on debug config
+ *
+ * `config.debug.creepLog.roles` and `config.debug.creepLog.rooms` define
+ * logging on common methods
+ *
+ * @param messages The message to log
+ */
+Creep.prototype.creepLog = function(...messages) {
+  if (config.debug.creepLog.roles.indexOf(this.memory.role) < 0) {
+    return;
+  }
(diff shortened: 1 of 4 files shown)


src/prototype_creep.js
```diff
@@ -42,6 +42,7 @@ Creep.prototype.moveToMy = function(target, range) {
     visualizer.showSearch(search);
   }
 
+  this.creepLog('moveToMy search:' + JSON.stringify(search));
```
mschultheiss83: this.creepLog('moveToMy search:', JSON.stringify(search));
// use join on string (+) or array of strings? (,)

src/logging.js
```diff
@@ -5,3 +5,13 @@ Room.prototype.log = function(...messages) {
...
 };
+
+Creep.prototype.creepLog = function(...messages) {
+  if (config.debug.creepLog.roles.indexOf(this.memory.role) < 0) {
+    return;
+  }
+  if (config.debug.creepLog.rooms.indexOf(this.room.name) < 0) {
+    return;
+  }
+  console.log(`${Game.time} ${this.room.name.rpad(' ', 6)} ${this.name.rpad(' ', 20)} ${this.pos} ${messages.join(' ')}`);
```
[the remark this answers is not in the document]
TooAngel: Good point, thx

src/prototype_creep.js
TooAngel: :+1 thx

TooAngel:
Surely we can improve logging even further, any third party solution in mind?

I think we can debug single creeps much better with this system, I'm tired of adding and removing specific log lines for specific rooms and roles and the `creepLog` helps a lot in that regard.

mschultheiss83:
same here logging i love just by changing in my local config debuging same roles in different rooms...

mschultheiss83:
and everywhere where a creepLog makes it into the master we could later change to the debug or verbose mode....

third party solution: only one... elasticsearch

PR #479, opened 2017-10-28
Refactor `prototype_creep_mineral` and transfer energy from terminal …

TooAngel:
…to storage

### DIFF
--- a/src/prototype_creep_mineral.js
+++ b/src/prototype_creep_mineral.js
@@ -1,5 +1,12 @@
 'use strict';
 
+/**
+ * Check is a given lab as enough mineral for reaction
+ *
+ * @param {object} lab - The lab to check.
+ * @param {string} mineralType - The mineral type to check.
+ * @return {boolean} if the lab has enough of the mineral
+ */
 Creep.prototype.checkLabEnoughMineral = function(lab, mineralType) {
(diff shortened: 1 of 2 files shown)


mschultheiss83:
- so i dont like https://github.com/TooAngel/screeps/pull/479/files#diff-eab562695f336fe618f2b1c287e7f4fcR439

can we move this to a config?

maybe if we could 
- move all the functions like tranfer and get to a namespace?

like `Creep.prototype.mineral.tranfer` and `Creep.prototype.mineral.get` and so on?
than in config we could use action: 'mineral.tranfer'.. 


so within a execute this should work to get the function:
 `_.get(creep, ['mineral.tranfer'])`

TooAngel:
Yes, no.

I agree this can be improved a lot.

My main goal was to introduce the functionality to transfer energy from the terminal to the storage if it exceeds the maximum. While touching the code I wanted to make a bit better: Moving all methods one up in the scope. Next time touching it, I will improve further, but currently I don't want to dig deeper, because I'm mostly eyeing on another topic, this was just to improve my rooms with a full terminal.

democratic-collaboration:
A new review, yeah.

            Votes: 250/551
            Coefficient: 0.453720508167
            Merging in 3 days 20 hours
            Age 1 days 14 hours

PR #490, opened 2017-11-30
Fix freezing when enemy enters room

Evanito:
indexOf() is for arrays and strings, apparently this object was neither.
For some reason, every name returns -1. 
Is friends just broken?


Evanito:
Issue was on my end, my friends.js said:
`global.exports =` instead of `module.exports =`

PR #510, opened 2018-06-13
Improve integration tests

### LINKED ISSUE #101: Stuck rooms on private server
- [x] W5N1: X10 Y9
 - [x] W8N3: 14/17

TooAngel:
- Use mongodb as screeps backend for tests
 - Use screepsmod-tickrate for test server
 - Remove `creepsNames` from test script, it is not used
 - Extend the players to test

Fixes #101


mschultheiss83:
do i understand this right, if i say you wannt to save the default your rooms in a mongo db, and every time we test, we load this db?

TooAngel:
I'm using the mongo mod, which uses MongoDB as the backend instead of the lokiJS database. This increases the performance significantly so that we should be able to test even longer sequences or have multiple bots running. Currently, I'm testing with 5 in parallel, PR will come soonish.

mschultheiss83:
no progress in no room?

TooAngel:
> no progress in any room?

Yeah on circleCI it is not yet robust, working on it.

TooAngel:
@mschultheiss83 There we go

utils/test_on_server.js
```diff
@@ -292,18 +306,25 @@ async function connectToCli() {
   socket.on('data', async (data) => {
     data = data.toString('utf8');
     const line = data.replace(/^< /, '').replace(/\n< /, '');
-    if (spawnBots(line, socket)) {
+    if (await spawnBots(line, socket)) {
       return;
     }
     if (setPassword(line, socket)) {
```
mschultheiss83: why not await here?
`await setPassword...` and `const setPassword = async function(line, socket) { ` on https://github.com/TooAngel/screeps/pull/510/files#diff-b532caf1840c952b1bfc47fa31a0119fR231

utils/test_on_server.js
```diff
@@ -26,11 +29,14 @@ for (let room of rooms) {
   status[room] = {
     controller: null,
     creeps: 0,
-    creepsNames: [],
+    // creepsNames: [],
```
mschultheiss83: remove or add debug flag same for https://github.com/TooAngel/screeps/pull/510/files#diff-b532caf1840c952b1bfc47fa31a0119fR147

utils/test_on_server.js
TooAngel: I need to change the `spawnBots` to an `async` function, to be able to use `await`. This is not necessary for `setPassword` (yet).

utils/test_on_server.js
TooAngel: Yeah, ... tbh I kept it like this because I think it will be reenabled soonish again. I have some more changes in the pipeline, will think about it and change it in an upcoming PR.

PR #517, opened 2018-06-28
Introduce `debugLog` for rooms

TooAngel:
`debugLog` logging can be configured via `config.debug` and simplifies
logging, debug and configuration which log should be shown.

### DIFF
--- a/src/logging.js
+++ b/src/logging.js
@@ -2,6 +2,12 @@ Room.prototype.log = function(...messages) {
   console.log(`${Game.time} ${this.name.rpad(' ', 27)} ${messages.join(' ')}`);
 };
 
+Room.prototype.debugLog = function(type, ...messages) {
+  if (config.debug[type]) {
+    this.log(messages);
+  }
+};
+
 RoomObject.prototype.log = function(...messages) {
   console.log(`${Game.time} ${this.room.name.rpad(' ', 6)} ${this.name.rpad(' ', 20)} ${this.pos} ${messages.join(' ')}`);
 };
--- a/src/prototype_creep.js
+++ b/src/prototype_creep.js
@@ -344,19 +344,23 @@ Creep.prototype.killPrevious = function(path) {
     return false;
   }
 
+  let killWho;
+  let killWhoName;
   if (this.ticksToLive < previous.ticksToLive) {
-    if (this.ticksToLive > 100) {
-      this.log('kill me: me: ' + this.ticksToLive + ' they: ' + previous.ticksToLive);
-    }
-    this.memory.killed = true;
-    this.suicide();
+    killWhoName = 'me';
+    killWho = this;
   } else {
-    if (previous.ticksToLive > 100) {
-      this.log('kill other: me: ' + this.ticksToLive + ' they: ' + previous.ticksToLive);
-    }
-    previous.memory.killed = true;
-    previous.suicide();
+    killWhoName = 'other';
+    killWho = previous;
   }
+
+  if (killWho.ticksToLive > killWho.memory.timeToTravel) {
+    this.log(`kill ${killWhoName} - me ttl: ${this.ticksToLive} they ttl: ${previous.ticksToLive}
+me: ${JSON.stringify(this)} Memory: ${JSON.stringify(this.memory)}
+other: ${JSON.stringify(previous)} Memory: ${JSON.stringify(previous.memory)}`);
+  }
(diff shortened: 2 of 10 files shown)


src/prototype_room_my.js
```diff
@@ -223,16 +223,18 @@ Room.prototype.checkNeedHelp = function() {
     if (!oldNeedHelp) {
       Memory.needEnergyRooms.push(this.name);
       this.memory.needHelp = true;
-      return '---!!!---' + this.name + ' need energy ---!!!---';
+      this.debugLog('energyTransfer', '---!!!---' + this.name + ' need energy ---!!!---');
+      return true;
     }
-    return 'Already set as needHelp';
+    return true;
```
mschultheiss83: let me do the return statements

src/prototype_room_my.js
```diff
@@ -293,11 +295,7 @@ Room.prototype.checkCanHelp = function() {
 Room.prototype.updateEnergyStatsAndCheckForHelp = function() {
   this.memory.energyStats.average = this.memory.energyStats.sum / this.memory.energyStats.ticks;
   const needHelp = this.checkNeedHelp();
-  if (needHelp) {
-    if (needHelp !== 'Already set as needHelp') {
-      this.log(needHelp);
-    }
-  } else {
+  if (!needHelp) {
```
mschultheiss83: same here, let me do

src/prototype_creep_move.js
```diff
@@ -54,9 +54,8 @@ Creep.prototype.moveToMy = function(target, range=1) {
 
   // Fallback to moveTo when the path is incomplete and the creep is only switching positions
   if (search.path.length < 2 && search.incomplete) {
-    this.creepLog(`moveToMy fallback ${JSON.stringify(target)} ${JSON.stringify(search)}`);
-    this.moveTo(target, {range: range});
-    return false;
+    this.room.debugLog('routing', `moveToMy fallback target: ${JSON.stringify(target)} range: ${range} search: ${JSON.stringify(search)}`);
```
mschultheiss83: `debuglog ` on RoomObjects

src/prototype_creep.js
```diff
@@ -344,19 +344,23 @@ Creep.prototype.killPrevious = function(path) {
...
-    if (previous.ticksToLive > 100) {
-      this.log('kill other: me: ' + this.ticksToLive + ' they: ' + previous.ticksToLive);
-    }
-    previous.memory.killed = true;
-    previous.suicide();
+    killWhoName = 'other';
+    killWho = previous;
+  }
+
+  if (killWho.ticksToLive > killWho.memory.timeToTravel) {
```
mschultheiss83: nice cleanup, but since when do we have a `creep.memory.timeToTravel`?

src/logging.js
```diff
@@ -2,6 +2,12 @@ Room.prototype.log = function(...messages) {
   console.log(`${Game.time} ${this.name.rpad(' ', 27)} ${messages.join(' ')}`);
 };
 
+Room.prototype.debugLog = function(type, ...messages) {
+  if (config.debug[type]) {
```
mschultheiss83: maybe add a `test flag` for circleci.com, like 
````
  if (config.debug[type] || config.debug.circleci) {
```
for loging everything?

mschultheiss83:
+1 for codeclimate fix

src/logging.js
TooAngel: Actually, I'm introducing this to reduce the noise to make exceptions or unhandled cases more prominent.

We could introduce a `config_local.js` for circleCi, for now I would like to reduce the log output.

src/prototype_creep.js
TooAngel: We have it for a long time, this indicates when the next creep need to be spawned e.g. for reserver, upgrader or sourcerer

src/prototype_creep_move.js
TooAngel: We have creepLog for creeps so that you can only get the log output for a single creep in a specific room. In this case I wanted it to be enabled for `routing` for debugging purpose.

src/prototype_room_my.js
TooAngel: Can revert this, to simplify it for you.

I think the changes make sense in this context, so they would belong here ...

src/prototype_room_my.js
TooAngel: Will revert, too.

PR #531, opened 2018-07-12
pickupEnergy with FIND_TOMBSTONES

mschultheiss83:


mschultheiss83:
@TooAngel please help on eslint:check FIND_TOMBSTONES

TooAngel:
Sure https://github.com/PostCrafter/eslint-plugin-screeps/pull/4

src/prototype_creep_resources.js
```diff
@@ -280,6 +280,26 @@ Creep.prototype.pickupEnergy = function() {
     }
   }
 
+  // FIND_TOMBSTONES and get them empty first
```
TooAngel: We should create a new function for it `Creep.prototype.withdrawTombstone` and call it from the `getEnergy()` function.

src/prototype_creep_resources.js
```diff
@@ -280,6 +280,26 @@ Creep.prototype.pickupEnergy = function() {
...
+  const tombstones = this.pos.findInRange(FIND_TOMBSTONES, 1, {
+    filter: (t) => {
+      return t.creep.my;
```
TooAngel: I think you can also withdraw from others, so you don't need to filter.

src/prototype_creep_resources.js
```diff
@@ -280,6 +280,26 @@ Creep.prototype.pickupEnergy = function() {
...
+    },
+  });
+  if (tombstones.length > 0) {
+    const creep = this;
+    _.map(tombstones, (t) => {
+      if (t.store[RESOURCE_ENERGY] > 0) {
```
TooAngel: Why only energy?

src/prototype_creep_resources.js
mschultheiss83: good point

src/prototype_creep_resources.js
mschultheiss83: i will do

src/prototype_creep_resources.js
mschultheiss83: it was in the getEnergy function... but a onlyEnergy flag i would like, or why would a sourcer on his way pick up resources?

TooAngel: Yeah, good point

mschultheiss83:
can we merge this?

mschultheiss83:
so with the ´withdrawResourcesFromTargets´ the end of `role.storagefiller.action` will be better to read

democratic-collaboration:
A new review, yeah.

            Votes: 52/617
            Coefficient: 0.08427876823338736
            Merging in 15 days 23.630277777777778 hours
            Age 2 days 7.915555555555556 hours

PR #533, opened 2018-07-12
some improvement part 4

mschultheiss83:
added pickupEnergy in getEnergy
better sellByOthersOrders
fixed canHelpRooms + needEnergyRooms
fixed energy and power transfer of role_storagefiller

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -86,7 +86,7 @@ global.config = {
   },
 
   tower: {
-    healMyCreeps: false,
+    healMyCreeps: true,
     repairStructures: false,
   },
 
@@ -112,7 +112,7 @@ global.config = {
     maxRooms: 8,
     cpuPerRoom: 13, // Necessary CPU per room, prevent claiming new rooms
     revive: true,
-    maxDistance: 17,
+    maxDistance: 10,
     minNewRoomDistance: 2,
     minEnergyForActive: 1000,
     minDowngradPercent: 90,
@@ -230,8 +230,8 @@ global.config = {
   },
 
   terminal: {
-    minEnergyAmount: 80000,
-    maxEnergyAmount: 120000,
+    minEnergyAmount: 40000,
+    maxEnergyAmount: 50000,
     storageMinEnergyAmount: 20000,
   },
 
@@ -242,7 +242,7 @@ global.config = {
   },
 
   market: {
-    minAmountToSell: 100000,
+    minAmountToSell: 50000,
     minSellPrice: 0.6,
     energyCreditEquivalent: 1,
     sellByOwnOrders: true,
@@ -258,10 +258,10 @@ global.config = {
     buyPower: false,
     // 3M credits
     minCredits: 3000000,
-    // set activate buyPower your need to set a roomName here e.g. 'W2S6'
-    buyPowerRoom: false,
     // disable to use power only in gathered room
     sendPowerOwnRoom: true,
+    // equalizes the energy beween your rooms via termial
+    sendEnergyToMyRooms: true,
   },
 
   priorityQueue: {
--- a/src/config.js
+++ b/src/config.js
@@ -112,6 +112,8 @@ global.config = {
     maxRooms: 8,
     cpuPerRoom: 13, // Necessary CPU per room, prevent claiming new rooms
     revive: true,
+    // creep max run distance for next room
+    // if terminal should send energy rooms should be close
     maxDistance: 10,
     minNewRoomDistance: 2,
     minEnergyForActive: 1000,
@@ -230,6 +232,7 @@ global.config = {
   },
 
   terminal: {
+    // terminals should not have to much enrgy, but not to less
     minEnergyAmount: 40000,
     maxEnergyAmount: 50000,
     storageMinEnergyAmount: 20000,
@@ -242,6 +245,7 @@ global.config = {
   },
 
   market: {
+    // sets mineral in terminal could be called minAmountMinerlasNotToSell
     minAmountToSell: 50000,
     minSellPrice: 0.6,
     energyCreditEquivalent: 1,
--- a/src/prototype_creep_startup_tasks.js
+++ b/src/prototype_creep_startup_tasks.js
@@ -117,7 +117,6 @@ Creep.recycleCreep = function(creep) {
 };
 
 Creep.getEnergy = function(creep) {
-  creep.pickupEnergy();
   return creep.getEnergy();
 };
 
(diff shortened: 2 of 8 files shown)


src/config.js
```diff
@@ -112,7 +112,7 @@ global.config = {
     maxRooms: 8,
     cpuPerRoom: 13, // Necessary CPU per room, prevent claiming new rooms
     revive: true,
-    maxDistance: 17,
+    maxDistance: 10,
```
TooAngel: Why?

src/config.js
```diff
@@ -230,8 +230,8 @@ global.config = {
   },
 
   terminal: {
-    minEnergyAmount: 80000,
-    maxEnergyAmount: 120000,
+    minEnergyAmount: 40000,
+    maxEnergyAmount: 50000,
```
TooAngel: Why?

src/config.js
```diff
@@ -242,7 +242,7 @@ global.config = {
   },
 
   market: {
-    minAmountToSell: 100000,
+    minAmountToSell: 50000,
```
TooAngel: Same here, I think we should explain why we set a specific value / change to a specific value in the config.

src/prototype_creep_startup_tasks.js
```diff
@@ -117,6 +117,7 @@ Creep.recycleCreep = function(creep) {
 };
 
 Creep.getEnergy = function(creep) {
+  creep.pickupEnergy();
```
TooAngel: If we want to add that, it should be in `Creep.prototype.getEnergy` and not hidden in the abstraction.

src/prototype_room_market.js
```diff
@@ -169,15 +175,41 @@ Room.prototype.sendPowerOwnRooms = function() {
   }
 };
 
+Room.prototype.sendEnergyToMyRooms = function() {
+  if ((_.size(Memory.needEnergyRooms) > 0) && (_.size(Memory.canHelpRooms) > 0) &&
+    _.includes(Memory.canHelpRooms, this.name) && !_.includes(Memory.needEnergyRooms, this.name) &&
+    (this.terminal.store[RESOURCE_ENERGY] > config.terminal.minEnergyAmount)) {
+    const diffNeedEnergyRooms = _.difference(Memory.needEnergyRooms, Memory.canHelpRooms);
+    const needEnergyRooms = _.size(diffNeedEnergyRooms) > 0 ? diffNeedEnergyRooms : Memory.needEnergyRooms;
+    const myRoom = _.shuffle(needEnergyRooms)[0];
```
TooAngel: Why not send to the closest one? It's cheaper

src/prototype_room_market.js
```diff
@@ -169,15 +175,41 @@ Room.prototype.sendPowerOwnRooms = function() {
   }
 };
 
+Room.prototype.sendEnergyToMyRooms = function() {
+  if ((_.size(Memory.needEnergyRooms) > 0) && (_.size(Memory.canHelpRooms) > 0) &&
+    _.includes(Memory.canHelpRooms, this.name) && !_.includes(Memory.needEnergyRooms, this.name) &&
+    (this.terminal.store[RESOURCE_ENERGY] > config.terminal.minEnergyAmount)) {
+    const diffNeedEnergyRooms = _.difference(Memory.needEnergyRooms, Memory.canHelpRooms);
+    const needEnergyRooms = _.size(diffNeedEnergyRooms) > 0 ? diffNeedEnergyRooms : Memory.needEnergyRooms;
```
TooAngel: To be honest I wouldn't to that. We should rely on, that `Memory.needEnergyRooms` and the ones which actually need energy. (Add they shouldn't be in the other list)

src/prototype_room_market.js
```diff
@@ -169,15 +175,41 @@ Room.prototype.sendPowerOwnRooms = function() {
...
+        }
+      }
+    }
+  }
+  return false;
+};
+
 Room.prototype.handleMarket = function() {
-  if (!this.terminal) {
+  if (!this.terminal || this.terminal.cooldown || this.terminal.cooldown > 0) {
```
TooAngel: `if (!this.terminal || this.terminal.cooldown)` should be fine, if `this.terminal.cooldown` is 0, it is a `false`.

src/config.js
mschultheiss83: my terminal we full with energy at 80k+ and i had no space

src/config.js
mschultheiss83: same here, my terminal we full with energy at 80k+ and i had no space

mschultheiss83: no only energy at 😄

src/config.js
mschultheiss83: my creepy died before they reach 17, so why seach for far away places, that aren't rechable

src/prototype_creep_startup_tasks.js
mschultheiss83: it will be removed in next PR. `the hidden abstraction` we dont need. 
i dont like `Creep.getEnergy` and that `method.push` stuff.

src/prototype_room_market.js
mschultheiss83: i did that before, but than not all rooms get the energy they need, only the closest to the best farming, and i wannted to push as much in the controllers of all rooms as i could

src/prototype_room_market.js
mschultheiss83: you mean `!_.includes(Memory.needEnergyRooms, this.name) &&`
as i was working on the improvements, it did happen some times... i will remove

src/prototype_room_market.js
mschultheiss83: sure, changed

src/config.js
TooAngel: Mind to write this as a comment to the line, so that we know in the future why we went in this direction?

src/config.js
TooAngel: Mind to write this as a comment to the line, so that we know in the future why we went in this direction?

src/config.js
TooAngel: Mind to write this as a comment to the line, so that we know in the future why we went in this direction?

src/prototype_creep_startup_tasks.js
TooAngel: hm, if you touch the PR anyway, I would suggest to remove it, if it get's anyway removed.

mschultheiss83:
should merged after https://github.com/TooAngel/screeps/pull/532

src/prototype_creep_startup_tasks.js
```diff
@@ -125,18 +125,13 @@ Creep.repairStructure = function(creep) {
 };
 
 Creep.prototype.getEnergyFromHostileStructures = function() {
-  if (this.carry.energy) {
-    return false;
-  }
```
TooAngel: I would keep it like this. If the creep has energy we spare a `find()` call = cpu usage

democratic-collaboration:
A new review, yeah.

            Votes: 52/617
            Coefficient: 0.08427876823338736
            Merging in 12 days 0.17444444444444446 hours
            Age 6 days 7.371388888888889 hours

mschultheiss83:
please doubble check

mschultheiss83:
Added WIP until FIND_TOMBSTONES gets fixed
```
   src/prototype_creep_resources.js
    299 |  const tombstones = this.pos.findInRange(FIND_TOMBSTONES, 1);
                                                   ^ 'FIND_TOMBSTONES' is not defined.
```

PR #535, opened 2018-07-13
spawn no scouts if we have an observer

mschultheiss83:


src/prototype_room_my.js
```diff
@@ -193,9 +193,11 @@ Room.prototype.handleScout = function() {
   if (this.name === 'sim') {
     return false;
   }
+  const observers = this.findPropertyFilter(FIND_MY_STRUCTURES, 'structureType', [STRUCTURE_OBSERVER]);
+  const levelCheck = (this.controller.level >= 2 && this.controller.level < 8) || (this.controller.level === 8 && observers.length === 0);
```
TooAngel: I think it is the same as `const levelCheck = this.controller.level >= 2 && observers.length === 0`
Or to save the cpu: `const levelCheck = this.controller.level >= 2 && (this.controller.level < 8 || this.findPropertyFilter(FIND_MY_STRUCTURES, 'structureType', [STRUCTURE_OBSERVER]).length === 0)` because the `find` would only be executed when `this.controller.level === 8`.

PR #540, opened 2018-07-18
Remove jscs

TooAngel:


mschultheiss83:
why?

TooAngel:
jscs is deprecated, the grunt library has dependencies with security issues and it will not be fixed.

PR #578, opened 2020-07-27
Remove noise, some fixes

TooAngel:
- Update dependencies
- Exclude scout from `Not in Game.creeps` (unexpected dead creep)
message. Scouts which are handled by hostile creeps needs to be handled
better, but is also not an issue.
- Cleanup the if clause in `prototype_creep` a bit
- Return `moveToMy` result in `prototype_creep_fight` because it is
already a boolean


TooAngel:

PR #581, opened 2020-08-01
Added mainLocal and bugfixes.

Kill4Free:
Added mainLocal for custom code.
mainLocal example for Pixel generation.
Always fix misplacedSpawn.
Fixed cpuStats.enabled typo.
Fixed linkStorage search function.
If room energyAverage <= 300 spawn low harvester to get everything running.

### DIFF
--- a/config_local.js.example
+++ b/config_local.js.example
@@ -5,3 +5,12 @@
 config.stats.enabled = true;
 config.stats.summary = true;
 config.profiler.enabled = false;
+
+//--- Enables a summary with all rooms for every tick
+//global.config.ticksummary.bucket = true;
+//global.config.ticksummary.gcl = true;
+//global.config.ticksummary.room = true;
+//global.config.ticksummary.seperator = true; // Adds a seperatorline after each tick
+
+//---  Enables pixel generation for live server.
+//global.config.pixel.enabled = true;
--- a/src/main.js
+++ b/src/main.js
@@ -27,5 +27,18 @@ module.exports.loop = function() {
       brain.main.execute();
     }
   }
+  if (global.config.pixel.enabled) {
+    if (typeof PIXEL !== 'undefined') {
+      if (Game.cpu.bucket >= PIXEL_CPU_COST + global.config.pixel.minBucketAfter) {
+        Game.cpu.generatePixel();
+      }
+    }
+  }
   brain.stats.updateCpuStats();
+  try {
+    const mainLocal = require('./main_local');
+    mainLocal();
+  } catch (e) {
+    // empty
+  }
 };
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -451,7 +451,7 @@ Room.prototype.checkForPlaner = function() {
         amount = 6;
       }
     }
-    if (this.controller.level === 4 && this.memory.misplacedSpawn) {
+    if (this.controller.level >= 4 && this.memory.misplacedSpawn) {
       amount = 3;
     }
     this.checkRoleToSpawn('planer', amount);
--- a/src/role_harvester.js
+++ b/src/role_harvester.js
@@ -31,7 +31,20 @@ roles.harvester.settings = {
   maxLayoutAmount: 6,
 };
 roles.harvester.updateSettings = function(room) {
-  if (room.storage && room.storage.my && room.storage.store.energy > config.creep.energyFromStorageThreshold && room.energyAvailable > 350 && !room.memory.misplacedSpawn) {
+  // If the energy average is lower than 350 for a longer period at least one low level harvester will be spawned.
+  // Fix for the following problem:
+  // - no room is able to send a nextroomer
+  // - all creeps are dead
+  // - energyAvailable is higher than 300 but lower than 350 for a longer time
+  if (room.memory.energyStats.average <= 350) {
+    // Layoutcost minimum: layout 250 -> 250
+    return {
+      layoutString: 'MWC',
+      amount: [2, 1, 1],
+      maxLayoutAmount: 6,
+    };
+  } else if (room.storage && room.storage.my && room.storage.store.energy > config.creep.energyFromStorageThreshold && room.energyAvailable > 350 && !room.memory.misplacedSpawn) {
+    // Layoutcost minimum: prefix 250 + layout 100 -> 350
     return {
       prefixString: 'WMC',
       layoutString: 'MC',
--- a/src/utils.js
+++ b/src/utils.js
@@ -221,4 +221,8 @@ global.utils = {
     return returnValue;
   },
 
+  // return: String with fixed width for rounded number
+  lpadround: function(nr, lpad, digest) {
+    return nr.toFixed(digest).padStart(lpad + digest + 1);
+  },
 };
--- a/main_local.js.example
+++ b/main_local.js.example
@@ -1,0 +1,7 @@
+module.exports = function() {
+  try{
+    console.log('mainLocal');
+  } catch (e) {
+    console.log(e.stack);
+  }
+}
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,2 +1,3 @@
 config_local.js
+main_local.js
 friends.js
--- a/src/config.js
+++ b/src/config.js
@@ -87,6 +87,9 @@ global.config = {
     constructionSites: false,
     routing: false,
     brain: false,
+    // Check bugs:
+    // - check wrong link order
+    checkbugs: false,
   },
 
   tower: {
@@ -138,6 +141,18 @@ global.config = {
     disabled: false,
     energyForCreeps: 800000,
     energyForSpawn: 250000,
+  },
+
+  pixel: {
+    enabled: false,
+    minBucketAfter: 2500,
+  },
+
+  ticksummary: {
+    bucket: false,
+    gcl: false,
+    room: false,
+    seperator: false,
   },
 
   buildRoad: {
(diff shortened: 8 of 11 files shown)


config_local.js.example
```diff
@@ -5,3 +5,11 @@
 config.stats.enabled = true;
 config.stats.summary = true;
 config.profiler.enabled = false;
+
+global.mainLocal = function() {
```
TooAngel, line 9: I wouldn't put logic into the config files.
IMHO the logic can be moved to `src/main.js` after `brain.stats.updateCpuStats();`, with a feature flag in the config `config.pixel.enabled`.

Instead of checking the bucked you could check for CPU still available after the rest of the tick was already executed.

src/prototype_room_my.js
```diff
@@ -54,7 +54,16 @@ Room.prototype.getLinkStorage = function() {
       return link;
     }
   }
-  const linkPos = this.memory.position.structure.link[0];
+
+  let linkPos = null;
```
TooAngel, line 58: In the room setup `prototype_room_basebuilder` the first link is placed next to the filler, also the link transfer logic always transfer the energy to the first link.

If this does not work out, we should check what happened on the initial room setup.

If this helps for your rooms, you could also just add a log output in case the first link is not the one desired, something like: 'First link is not the best for filler, please manually check room setup'. So that someone ./ me can look into it, in case it happens for me, too.

src/role_harvester.js
```diff
@@ -31,7 +31,13 @@ roles.harvester.settings = {
   maxLayoutAmount: 6,
 };
 roles.harvester.updateSettings = function(room) {
-  if (room.storage && room.storage.my && room.storage.store.energy > config.creep.energyFromStorageThreshold && room.energyAvailable > 350 && !room.memory.misplacedSpawn) {
+  if (room.memory.energyStats.average <= 300) {
```
TooAngel, line 34: I would guess that this should be handled already in the harvester config. I didn't have such cases, but I'm fine to be extra safe.

config_local.js.example
Kill4Free: Thx for your feedback.

I think you're right the config shouldn't include code!
I've got some custom logic that prints links to the rooms with stats and the current queue. This way I've got an overview over all rooms in the console. That's why I would like to have an entrypoint into the main function.
What do you think about a `main_local.js` instead of the mainLocal from `config_local.js`?

The current logic for pixel generation could be extended with another condition to check if there's still CPU left and with an config option to enable pixel generation.
I think thats a good idea. The pixel generation is just using the bucket and should include PIXEL_CPU_COST + 1000 so that there is always at least 1k cpu left.

src/prototype_room_my.js
Kill4Free: I've checked the code currently the `updatePosition` from `prototype_room_init.js` [Link](https://github.com/TooAngel/screeps/blob/e99584fddd04d97baa7eae1f089a50a1b5d7363e/src/prototype_room_init.js#L128-L135) firstly adds the links for the sources inside `initSetSources()` afterwards the storage link is generated inside `setFillerArea()`.
I think this should be fixed and my check should be included for a log statement.

src/role_harvester.js
Kill4Free: I checked the harvester config and couldn't find where it's handled.

src/prototype_room_my.js
Kill4Free: I just checked my rooms and all rooms have this wrong link ordering.

config_local.js.example
TooAngel: Regarding: `main_local`: If it helps, include it.
But, if you have logging or logical which helps, why not include it directly. E.g. there are the `config.debug` options that make the logging more verbose. If you need additional information, make a config flag, default to `false`. In your local config enable it. This way others can use that, too.

regarding pixel: Sounds good,

src/prototype_room_my.js
TooAngel: Ok, thx, that 'bug' was introduced ... by me :-) I opened a new issue for it #582 

So if you could include a logline here, that would be cool. Let's fix it in another PR.

src/role_harvester.js
TooAngel: Hm, what are you trying to accomplish?

You are checking for `energyStats.average <= 300`, the current code checks for `energyAvailable > 350`. What are the cases where this runs into issues?
Would it make sense to change the `energyAvailable` to `energyStats.average`, so that we don't have a similar special case?!

Kill4Free: I had the following problem:
- if no room is able to send a nextroomer
- and all creeps are dead
- energyAvailable is higher than 300
no creep has been spawned. This way if the room energyAvailable is below 300 (maybe this should be set higher) there will be at least one low level harvester to get some energy.
I think the problem was, that the `prefixString: 'WMC'` with the normal layout needs too much energy.

src/prototype_room_my.js
Kill4Free: Okay I will add the logline in another spot and remove the new code. 👍 
Maybe it would be good to add an option `config.debug.checkbugs` so that this check is not executed always?

config_local.js.example
Kill4Free: Added configuration for `pixels` and `tickSummary` in latest commit.
Added `main_local` in latest commit.

Any suggestions regarding the latest changes?

src/utils.js
```diff
@@ -221,4 +221,7 @@ global.utils = {
     return returnValue;
   },
 
+  lpadround: function lpadround(nr, lpad, digest) {
```
TooAngel, line 224: I recently discovered `padStart` and `padEnd` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)

Wondering if this wouldn't be easier to read. So something like `return nr.padStart(digest+1pad+1).padEnd(digest+1pad+2)` (Not sure if this matches exactly what you want to accomplish).

config_local.js.example
TooAngel: Add some comments, basically I'm fine.

src/prototype_room_my.js
TooAngel: Yeah, sounds good

src/role_harvester.js
TooAngel: Hm, `WMC` should be 200 energy, so buildable. Not sure if the creep configurator messed this up.
Can you add exactly this scenario / text as a comment to the if clause, so that we are aware when looking into it again?

src/main.js
```diff
@@ -27,5 +27,18 @@ module.exports.loop = function() {
...
+    if (typeof PIXEL !== 'undefined') {
+      if (Game.cpu.bucket >= PIXEL_CPU_COST + global.config.pixel.minBucketAfter) {
+        Game.cpu.generatePixel();
+      }
+    }
+  }
   brain.stats.updateCpuStats();
+  try {
+    require('./main_local');
+    mainLocal();
```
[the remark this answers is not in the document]
Kill4Free: There's no advantage I think. Just haven't thought about it

src/utils.js
Kill4Free: It's a round function with leading spaces and fixed digest count.

src/role_harvester.js
Kill4Free: I will add this as a comment to the if clause.

config_local.js.example
Kill4Free: I will polish this the next few days then we can merge it 👍

src/utils.js
TooAngel: Ah, missed somehow the rounding, sry.

How about: `return nr.toFixed(digest).padStart(lpad + digest);`

config_local.js.example
TooAngel: Sounds good, can you please squash your changes to a single commit when you are satisfied?!

src/role_harvester.js
Kill4Free: If I'm not wrong the problem seems to be the prefix + layout which is 350 together.

src/utils.js
Kill4Free: I changed it to your suggestion it's easier understandable.

src/role_harvester.js
TooAngel: K, created an issue: #585

Kill4Free:
I've added the comments, updated the other code and squashed all commits into one.
Do you have any further comments?

src/role_harvester.js
```diff
@@ -31,7 +31,20 @@ roles.harvester.settings = {
...
+  // - energyAvailable is higher than 300 but lower than 350 for a longer time
+  if (room.memory.energyStats.average <= 350) {
+    // Layoutcost minimum: layout 250 -> 250
+    return {
+      layoutString: 'MWC',
+      amount: [2, 1, 1],
+      maxLayoutAmount: 6,
+    };
+  } else if (room.storage && room.storage.my && room.storage.store.energy > config.creep.energyFromStorageThreshold && room.energyAvailable > 350 && !room.memory.misplacedSpawn) {
+    // Layoutcost minimum: prefix 250 + layout 100 -> 350
```
TooAngel, line 47: Fine, let's keep it like this.

But, the if clause explicitly states, that `room.energyAvailable` needs to be `> 350` otherwise the default settings are used.

TooAngel:
The pull request will be automatically merged in ~5 days.

src/role_harvester.js
Kill4Free: Mhh you're right but I had this problem sometimes. I will check if this problem persists!

TooAngel:

PR #587, opened 2020-08-15
Fix creep spawn direction

### LINKED ISSUE #584: Use spawn direction
AFAIK the direction where a creep leaves a spawn was introduced a while ago. E.g. in room `E33N8` the wrong direction leads to problems.

The spawn direction (or directions) could be stored as an additional value in the `positions` for each spawn

TooAngel:
- `createCreep` is deprecated and replaced by `spawnCreep`
- The room layout makes sure the spawns are below the path, so fixing
the spawn direction to `TOP`. Further improvements can be to change the
room layout and store the spawn direction.
- Update dependencies

Fixes #584


TooAngel:
@Kill4Free would be nice if you can review the changes. I did some code cleanup on the new `checkBugs` implementation.

Kill4Free:
@TooAngel I'm fine with extracting `checkBugs` into another function.

PR #589, opened 2020-08-18
Fixes harvester bug for high gcl and energy <= 350

### LINKED ISSUE #585: Harvester problem with only 300 energy in room on higher RCL
It seems that there is an issue with harvesters: https://github.com/TooAngel/screeps/pull/581#discussion_r464626547

The best approach would be to write tests for different cases and see when it fails. Any other solution is fine, too :-)

Kill4Free:
Fixed creep builder updateSettings function. (lodash merge)
Harvester prefers empty spawns and extensions before anything else.
Added logging to config.debug.spawn.
Cleared unnecessary stuff from harvester.updateSettings.

Fixes #585

### DIFF
--- a/src/config_brain_stats.js
+++ b/src/config_brain_stats.js
@@ -62,6 +62,7 @@ brain.stats.add = function(path, newContent) {
     current = current[item];
   }
 
+  // Redundant _.merge automatically alters first object
   current = _.merge(current, newContent);
   return true;
 };
--- a/src/prototype_creep_resources.js
+++ b/src/prototype_creep_resources.js
@@ -726,9 +726,18 @@ Creep.prototype.construct = function() {
 };
 
 Creep.prototype.getTransferTargetStructure = function() {
-  const structure = this.pos.findClosestStructureWithMissingEnergyByRange(
+  let structure = this.pos.findClosestStructureWithMissingEnergyByRange(
     (object) => object.structureType !== STRUCTURE_STORAGE,
   );
+  // Harvester should always prefer Spawn and Extensions
+  if (this.memory.role === 'harvester') {
+    const structureSpawnExtension = this.pos.findClosestStructureWithMissingEnergyByRange(
+      (object) => (object.structureType === STRUCTURE_SPAWN || object.structureType === STRUCTURE_EXTENSION),
+    );
+    if (structureSpawnExtension !== null) {
+      structure = structureSpawnExtension;
+    }
+  }
   if (structure === null) {
     if (this.room.storage && this.room.storage.my && this.memory.role !== 'planer') {
       this.memory.target = this.room.storage.id;
--- a/src/prototype_room_creepbuilder.js
+++ b/src/prototype_room_creepbuilder.js
@@ -211,7 +211,14 @@ Room.prototype.getPartsStringDatas = function(parts, energyAvailable) {
 Room.prototype.getSettings = function(creep) {
   const role = creep.role;
   const updateSettings = roles[role].updateSettings && roles[role].updateSettings(this, creep);
-  const settings = _.merge(roles[role].settings, updateSettings);
+  if (config.debug.spawn) {
+    this.log(`getSettings updateSettings: ${JSON.stringify(updateSettings)}`);
+    this.log(`getSettings roles[role].settings: ${JSON.stringify(roles[role].settings)}`);
+  }
+  const settings = _.merge({}, roles[role].settings, updateSettings);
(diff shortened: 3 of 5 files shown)


TooAngel:
This pull request will be automatically merged by [worlddriven](https://www.worlddriven.org) in 9 days and 23 hours.
Check the `worlddriven` status checks or the [dashboard](https://www.worlddriven.org/TooAngel/screeps/pull/589) for actual stats.

`Approved` reviews will speed this up.
`Request Changes` reviews will slow it down or stop it.

src/config_brain_stats.js
```diff
@@ -62,6 +62,7 @@ brain.stats.add = function(path, newContent) {
     current = current[item];
   }
 
+  // Redundant _.merge automatically alters first object
```
TooAngel, line 65: If it is redundant, just fix it (thx, for pointing out)

src/prototype_creep_resources.js
```diff
@@ -726,9 +726,18 @@ Creep.prototype.construct = function() {
 };
 
 Creep.prototype.getTransferTargetStructure = function() {
-  const structure = this.pos.findClosestStructureWithMissingEnergyByRange(
+  let structure = this.pos.findClosestStructureWithMissingEnergyByRange(
     (object) => object.structureType !== STRUCTURE_STORAGE,
   );
+  // Harvester should always prefer Spawn and Extensions
```
TooAngel, line 732: Harvester only uses this method before the storage is built, so the only structures are `SPAWN`, `EXTENSION`, and `TOWER`. Because of invaders and early attacks, I think it is a good idea to have the `TOWER` filled, too.

As soon as the storage is built up (and has enough energy) it will move on a fixed path and only fills spawns and extensions. I could only see issues at that point when also other structures are available.

So fine leave it in right now, but needs to be verified in early scenarios regarding the tower.

src/prototype_room_creepbuilder.js
```diff
@@ -211,7 +211,14 @@ Room.prototype.getPartsStringDatas = function(parts, energyAvailable) {
 Room.prototype.getSettings = function(creep) {
   const role = creep.role;
   const updateSettings = roles[role].updateSettings && roles[role].updateSettings(this, creep);
-  const settings = _.merge(roles[role].settings, updateSettings);
+  if (config.debug.spawn) {
```
TooAngel, line 214: Instead of having the `if clause` here, you can use the `debugLog`, like:
```this.debugLog('spawn', `getSettings updateSettings: ${JSON.stringify(updateSettings)}`);``` (checkout `logging.js`)
which reduces the complexity in this methods and has the same functionality.

File: src/prototype_room_creepbuilder.js [hunk 4]
```diff
@@ -211,7 +211,14 @@ Room.prototype.getPartsStringDatas = function(parts, energyAvailable) {
...
+    this.log(`getSettings updateSettings: ${JSON.stringify(updateSettings)}`);
+    this.log(`getSettings roles[role].settings: ${JSON.stringify(roles[role].settings)}`);
+  }
+  const settings = _.merge({}, roles[role].settings, updateSettings);
```
TooAngel, line 218: Wow, nice catch, tricky to find, thx

TooAngel (approved):
Looks good, thx, nice catch.

Maybe if you change the debug logging to `Room.debugLog` we also get rid of the codeclimate issues.

src/config_brain_stats.js
Kill4Free: Okay will fix it if I've got some spare time!

src/prototype_creep_resources.js
Kill4Free: I had a problem while all buildings were done. There was just an harvester from the previous bug and this harvester only filled the `TERMINAL`.
The tower would be filled after the room is working smootly again don't know if that brings a problem in the early stages.

src/prototype_room_creepbuilder.js
Kill4Free: Oh I overlooked that I will change these accordingly.

src/prototype_room_creepbuilder.js
Kill4Free: Yeah took me some time to find it but in the end I found it :-)

src/prototype_creep_resources.js
TooAngel: Ok, let's leave it like this. I don't like special cases so much, but better be robust for now.

TooAngel:

Issue #542, opened 2018-07-23
startMeleeSquad() TypeError: Cannot read property '0' of undefined


LokiSharp:
After I use brain.startSquad('W54N11','W56N11')

[下午10:13:45][shard2]8076332 W54N11 squadsiege-3351      [room W54N11 pos 11,15] preMove
[下午10:13:45][shard2]8076332 W54N11 squadsiege-3351      [room W54N11 pos 11,15] Executing creep role failed: W54N11 squadsiege-3351 5b55dfa27358607bf5fd2b6c {"x":11,"y":15,"roomName":"W54N11"} TypeError: Cannot read property '0' of undefined
TypeError: Cannot read property '0' of undefined
    at Object.RoomPosition.getAdjacentPosition (prototype_roomPosition:78:58)
    at Object.roles.squadsiege.dismantleSurroundingStructures (role_squadsiege:21:32)
    at Object.roles.squadsiege.preMove (role_squadsiege:42:20)
    at Object.Creep.followPath (prototype_creep_routing:149:42)
    at Object.Creep.handle (prototype_creep:107:14)
    at Object.Room.execute (prototype_room:26:13)
    at brain.main.roomFilter (config_brain_main:21:7)
    at :17946:13
    at :19087:15
    at baseForOwn (:18060:14)
[下午10:13:49][shard2]8076333 W54N11 squadsiege-3351      [room W54N11 pos 11,15] preMove
[下午10:13:49][shard2]8076333 W54N11 squadsiege-3351      [room W54N11 pos 11,15] Executing creep role failed: W54N11 squadsiege-3351 5b55dfa27358607bf5fd2b6c {"x":11,"y":15,"roomName":"W54N11"} TypeError: Cannot read property '0' of undefined
TypeError: Cannot read property '0' of undefined
    at Object.RoomPosition.getAdjacentPosition (prototype_roomPosition:78:58)
    at Object.roles.squadsiege.dismantleSurroundingStructures (role_squadsiege:21:32)
    at Object.roles.squadsiege.preMove (role_squadsiege:42:20)
    at Object.Creep.followPath (prototype_creep_routing:149:42)
    at Object.Creep.handle (prototype_creep:107:14)
    at Object.Room.execute (prototype_room:26:13)
    at brain.main.roomFilter (config_brain_main:21:7)
    at :17946:13
    at :19087:15
    at baseForOwn (:18060:14)
[...]

PR #609, opened 2020-09-22
Fixed siege squad, removed some useless log

Gicminos:
Fixes https://github.com/TooAngel/screeps/issues/542

### DIFF
--- a/src/prototype_creep_fight.js
+++ b/src/prototype_creep_fight.js
@@ -198,10 +198,10 @@ Creep.prototype.siege = function() {
     Game.notify(Game.time + ' ' + this.room.name + ' Attacking');
     this.memory.notified = true;
   }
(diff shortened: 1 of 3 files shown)


TooAngel (approved):
Thanks for contributing.

A bit more detailed commit/Pull Request description would be nice, it's okay.

TooAngel (approved):

Gicminos:
As said, I was getting the same problem happening in https://github.com/TooAngel/screeps/issues/542 and I managed to solve it.

At the same time I had one more problem with https://github.com/TooAngel/screeps/blob/0bb9f632ce27c0bc27c22d2901226238906b9e30/src/prototype_creep_fight.js#L201 were the siege minion was not able to find the rampant (old function returning empty result even if the room had towers and spawn). And with that correction I fixed that too.

TooAngel (approved):

TooAngel:

PR #620, opened 2020-11-05
Bugfix nextroomers not being called and creep builder

connorzg:
Have been playing with this bot for the last few days and noticed/fixed a couple bugs that popped up. 

- The creepbuilder was frequently returning `spawnCreateCreep body too long` because the `getPartConfigMaxBodyLength` was not properly being passed the prefix and suffix strings

- Nextroomers were not being called, nor rooms revived because `reviveMyNowHelperValid` previously never returned true, meaning the following code block would always execute and it never got to the `helperRoom.checkRoleToSpawn('nextroomer', ...)` line

```
    if (!this.reviveMyNowHelperValid(helperRoom)) {
      continue;
    }
```

### DIFF
--- a/src/prototype_room_creepbuilder.js
+++ b/src/prototype_room_creepbuilder.js
@@ -302,7 +302,7 @@ Room.prototype.getPartConfig = function(creep) {
     sufixString,
     fillTough} = settings;
   let layoutString = settings.layoutString;
-  const maxBodyLength = this.getPartConfigMaxBodyLength();
+  const maxBodyLength = this.getPartConfigMaxBodyLength(prefixString, sufixString);
 
   const prefix = this.getPartsStringDatas(prefixString, energyAvailable);
   if (prefix.fail) {
--- a/src/prototype_room_my.js
+++ b/src/prototype_room_my.js
@@ -558,28 +558,23 @@ const checkForRoute = function(room, roomOther) {
 };
 
 Room.prototype.reviveMyNowHelperValid = function(helperRoom) {
+  let valid = false;
+
   if (helperRoom.isStruggeling()) {
     this.debugLog('revive', `No nextroomer is struggeling ${helperRoom.name}`);
-    return false;
-  }
-  // When close before downgrading, just send nextroomers
-  // TODO replace 1500 with CREEP_LIFE_TIME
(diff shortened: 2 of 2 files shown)


TooAngel (approved):
Nice catch, thank you for contributing, looks good.

connorzg:
@TooAngel thanks for the fast review! I'm just starting to dip my toes in this project but see multiple areas that need work so I'm excited to be able to contribute. The next thing I've been working on personally and will likely PR is improvements to the auto-move of misplaced spawns, I have seen that kill out many colonies because the workers die before the new spawn is completed. 

Increasing the number of planers and energy requirements has been a good first step but likely better would be to calculate the existing planners TTL and WORK components vs the work needed to build a spawn.

TooAngel:
Cool, happy to hear that you like it and you are interested in contributing, welcome.

I'm wondering that replacement of the misplaced spawn leads to many problems, I tested in on multiple rooms. But it can definitely be improved. If you know a room where that happens we could include it in the tests.

I'm trying to find simple maybe not accurate solutions, so before calculating the exact amount, which I guess is even hard to calculate, I would try the just put noteworthy higher thresholds than expected.

But go ahead show us your ideas :-)

connorzg:
@TooAngel I've been playing on a personal private server running my modified tooangel bot against some of the standard tooangel bots and have noticed close to half of the bots die out during the spawn reconstruction phase, though that could also be related to the nextroomers not being called (which this PR fixes!). I haven't checked out the tests you have in this repo yet so I'll see if that's something I can reproduce there.

That's a good point about complexity. This is the most cpu-constrained project I've worked on so I'll have to keep in mind simple is better, and try not to overengineer.

TooAngel:
The misplaced spawn shouldn't have to do with the nextroomers. The misplaced spawn only occurs in the first room if the spawn does not match the room layout. In other rooms, the spawn is placed automatically. Nextroomers are only spawned for additional rooms.

I just sent a PR #621 to update the package version, this will automatically release the current version to npm and steam so that you can play against the newest one.

Private server is perfect, if you can post the room name and spawn position (if you still have them), I can have a look, too.

TooAngel:

PR #622, opened 2020-11-24
Use named methods to find hostile structures in ranged range

TooAngel:


TooAngel:

PR #641, opened 2022-02-10
Thank _GrimReaper

TooAngel:
The public TooAngel account was wiped by other players, `_GrimReaper`
was so nice to build up a spawn again. As the reputation system does not
cover this yet, this commit will try to give at least something back.

Also
- Only `myRooms` should use `memory`.
- Small quest fixes
- Uninitialized object or array fixes
- Automated deploy only for the public world


TooAngel:

PR #645, opened 2022-04-04
[eslint] change: space after function

mschultheiss83:
- updated package.json (minor version)
- eslint change: space after function


.eslintrc-base
```diff
@@ -71,6 +71,7 @@ rules:
   max-len: [2, 240]
   max-statements-per-line: 2
   require-jsdoc: 0
+  space-before-function-paren: ["error", "always"]
```
TooAngel, line 74: hm, actually I wouldn't change that.

TooAngel (changes requested):
I wouldn't to that style change

mschultheiss83:
closed see https://github.com/TooAngel/screeps/pull/648

PR #649, opened 2022-04-08
Decrease CI server test runtime

TooAngel:
External Pull Requests have a runtime of ~1 hour on circleCi, so the
test runtime needs to be below this.

Moving to e.g. github actions could increase this again.


mschultheiss83:
is it possible to save state as artifact and re-run it to extend tests?

TooAngel:
> is it possible to save state as artifact and re-run it to extend tests?

puh, yeah, I think this could be somehow possible. I feel it is a bit hacky - for now I would just reduce the time so that we have at least some tests which are required.
After that we could look into better solutions.

mschultheiss83 (approved):
Thank you

TooAngel:

PR #655, opened 2022-09-06
[LOGIC REFINE]only do destroyStructureSpawn after you have at least several Spawns now.

XenoAmess:
failure of XenoAmess, from https://screeps.com/a/#!/history/shard1/E49S27?t=42643500 to https://screeps.com/a/#!/history/shard1/E49S27?t=42645100 , shows it is really dangerous to destroy StructureSpawn when you have too few destroyStructureSpawn, so please only do it when you have another 2, at least you will not die like what XenoAmess suffered.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -217,6 +217,7 @@ global.config = {
 
   myRoom: {
     underAttackMinAttackTimer: 50,
+    leastSpawnsToRebuildStructureSpawn: 1,
   },
 
   room: {
--- a/src/prototype_room_init.js
+++ b/src/prototype_room_init.js
@@ -376,6 +376,10 @@ Room.prototype.checkForSpawnPosition = function(pos) {
  **/
 Room.prototype.checkForMisplacedSpawn = function() {
   const spawns = this.findMySpawns();
+  const spawnsCount = spawns.length;
+  if (spawnsCount < config.myRoom.leastSpawnsToRebuildStructureSpawn) {
+    return;
+  }
   for (const spawn of spawns) {
     if (!this.checkForSpawnPosition(spawn.pos)) {
       this.log('Setting misplacedSpawn');
--- a/src/prototype_room_basebuilder.js
+++ b/src/prototype_room_basebuilder.js
@@ -102,7 +102,12 @@ Room.prototype.buildRampartsAroundSpawns = function() {
  * @return {boolean}
  */
 function destroyStructureSpawn(room, structure) {
+  const spawnsCount = room.findMySpawns().length;
   if (room.memory.misplacedSpawn) {
+    if (spawnsCount < config.myRoom.leastSpawnsToRebuildStructureSpawn) {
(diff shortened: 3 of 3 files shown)


CarsonBurke:
Why not just when there are 2+ spawns? Your issue only happened when there was 1 and it tried to be rebuilt

XenoAmess:
> Your issue only happened when there was 1 and it tried to be rebuilt

Yep, sadly the original codes be too confidence about the rebuilt, which comes to a failure.

> Why not just when there are 2+ spawns? 

because I still think 2 is not safe enough...

XenoAmess:
OK seems there is a really mean test there, so I just would set the default config to 1
I will strongly suggest make this config at least to 2, or even 3, but I will not force others.
:)

src/prototype_room_init.js
```diff
@@ -375,6 +375,10 @@ Room.prototype.checkForSpawnPosition = function(pos) {
  * @return {undefined}
  **/
 Room.prototype.checkForMisplacedSpawn = function() {
+  const spawnsCount = Object.keys(Game.spawns).length;
```
mschultheiss83, line 378: i dont think you should use Game.spawns as spawnsCount... check for 
```js
  const spawns = this.findMySpawns();
  if (spawns.length > 2) {
```

mschultheiss83 (changes requested):
maybe you should move this to
```js
// from
config.structureSpawn: {
  leastStructureSpawnToDestroyStructureSpawn
}
// to 
config.myRoom: {
   leastSpawnsToRebuildStructureSpawn
}
```

src/prototype_room_init.js
XenoAmess: @mschultheiss83 Done. please have a re-review.

src/config.js
```diff
@@ -217,6 +217,7 @@ global.config = {
 
   myRoom: {
     underAttackMinAttackTimer: 50,
+    leastSpawnsToRebuildStructureSpawn: 1,
```
mschultheiss83, line 220: so should this be 1 or 2

mschultheiss83 (approved):

src/config.js
XenoAmess: I would use 2 for myself, but there be a unit test here requires MUST relocate spawn as soon as possible, otherwise would fail.
So I have to set it default 1.

mschultheiss83: can you point me to it, can't find it?
image

XenoAmess: > can you point me to it, can't find it? image

I make it to 2 and commit, then we can see it from circleci result...

XenoAmess: @mschultheiss83 see https://app.circleci.com/pipelines/github/TooAngel/screeps/237/workflows/d2d2ba27-f834-4b4b-82f4-456636e8e5d8/jobs/2532

TooAngel:

PR #656, opened 2022-09-06
[LOGIC REFINE]when in a room which have your controller, but no your spawn, let builder be able to extract energy from everywhere, as it need to rebuild spawn at any cost.

XenoAmess:


src/prototype_creep_startup_tasks.js
```diff
@@ -133,6 +133,35 @@ Creep.repairStructure = function(creep) {
...
+    return false;
+  }
+  let structures = this.room.findPropertyFilter(FIND_STRUCTURES, 'structureType', [STRUCTURE_CONTROLLER, STRUCTURE_RAMPART, STRUCTURE_EXTRACTOR, STRUCTURE_OBSERVER], {
+    inverse: true,
+    filter: Room.structureHasEnergy,
+  });
+  if (this.carry.energy || !structures.length) {
+    return false;
+  }
+  // Get energy from the structure with highest amount first
```
mschultheiss83, line 150: why no the lowest?

XenoAmess: Because that means 2 things I want to avoid.
1. it would use Storage at last, and Storage have quite some energy and should be extracted out then use for the hard time.
2. lowest sometimes not worthy to walt to and extract, they just have too less energy. when a spawn is destroyed, I think we must rebuild it as fast as possible, so wasting time on the road seems not quite wise to me.

mschultheiss83: i don't know if i like that... getEnergyFromAnyStructures means even hostile storage
i sometimes find a left room let's say hostile, with an inactive player, so i claim the room.
what i don't want is that the nextroomers grab the energy from it. 

what i like more is that the universal grab the energy and build the extensions, towers and upgrade controller and when they reach the RoomControllerLevel 4 they need to rebuild the storage anyway

src/prototype_creep_resources.js
```diff
@@ -13,10 +13,12 @@ Creep.prototype.universalBeforeStorage = function() {
   if (this.room.controller && (this.room.controller.ticksToDowngrade < CONTROLLER_DOWNGRADE[this.room.controller.level] / 10 || this.room.controller.level === 1)) {
     methods.push(Creep.upgradeControllerTask);
   }
-  const universals = this.room.findPropertyFilter(FIND_MY_CREEPS, 'memory.role', ['universal']).map((creep) => creep.name);
-  universals.sort();
-  if (universals.indexOf(this.name) < 2) {
-    methods.push(Creep.transferEnergy);
+  if (!this.room.isConstructingSpawn()) {
```
mschultheiss83, line 16: why not transferEnergy if room.isConstructingSpawn?

mschultheiss83 (commented):
i understand you point, i would wish to disable this feature, maybe if some config is set like

```js
Room.prototype.isConstructingSpawn = function() {
  if (!this.controller || !this.controller.my || config.myRoom.enableGetEnergyFromAnyStructures) {
    return false;
  }
  if (this.findPropertyFilter(FIND_MY_STRUCTURES, 'structureType', [STRUCTURE_SPAWN], {}).length > 0) {
    return false;
  }
  return true;
};
```

src/prototype_creep_startup_tasks.js
XenoAmess: > i don't know if i like that... getEnergyFromAnyStructures means even hostile storage i sometimes find a left room let's say hostile, with an inactive player, so i claim the room. what i don't want is that the nextroomers grab the energy from it.
> 
> what i like more is that the universal grab the energy and build the extensions, towers and upgrade controller and when they reach the RoomControllerLevel 4 they need to rebuild the storage anyway

Come to think of it, It is actually OK for me to accept only grab energy from own structure, as it is hard to see any hostile structure in my waiting spawn room.

Will change it to only grab energy from mystructure.

src/prototype_creep_resources.js
XenoAmess: > maybe if some config is set like

OK, can do.

XenoAmess: > why not transferEnergy if room.isConstructingSpawn?

because transferEnergy have higher priority than constructTask.

means, even if you are rebuilding spawn, the universals will still fill your precious energy into extensions, which can be a disaster.

XenoAmess:
> i understand you point, i would wish to disable this feature, maybe if some config is set like
> 
> ```js
> Room.prototype.isConstructingSpawn = function() {
>   if (!this.controller || !this.controller.my || !config.myRoom.enableGetEnergyFromAnyStructures) {
>     return false;
>   }
>   if (this.findPropertyFilter(FIND_MY_STRUCTURES, 'structureType', [STRUCTURE_SPAWN], {}).length > 0) {
>     return false;
>   }
>   return true;
> };
> ```

Configuration added. please review when you have time.

Have a nice weekend.

mschultheiss83 (approved):

TooAngel:

PR #657, opened 2022-09-06
[SEVERE]Fix claimer code changed by wilken which cause can never claim any room when config.nextRoom.resourceStats be false, which be default...

XenoAmess:
@TooAngel I cannot say there be 0 possibility that it be a bug, but I really think it possibly be a hacking-attack to this repo. 

please review as soon as possible, thanks. 

only when Memory.myRooms.length < 0 can we claim new room, yep a negative array!

### DIFF
--- a/src/brain_memory.js
+++ b/src/brain_memory.js
@@ -202,11 +202,32 @@ Upgrade less: ${strings.upgradeLess}
 =========================`);
 };
 
+brain.executeEveryTicks = function(ticks) {
+  const timer = (ticks > 3000) ? Game.time - Memory.time + 1 : 0;
+  return (timer > 1) ? (Game.time % ticks) < timer : (Game.time % ticks) === 0;
+};
+
+brain.setDynamicConfigInternal = function() {
+  Memory.dynamicConfig = Memory.dynamicConfig || {};
+  Memory.dynamicConfig.nextRoom = Memory.dynamicConfig.nextRoom || {};
+  Memory.dynamicConfig.nextRoom.maxRooms = Math.min(config.nextRoom.maxRooms, Game.gcl.level);
+};
+
+brain.setDynamicConfigInternal();
+
+brain.setDynamicConfig = function() {
+  if (!this.executeEveryTicks(1000)) {
+    return false;
+  }
+  this.setDynamicConfigInternal();
+};
+
 brain.prepareMemory = function() {
   Memory.username = Memory.username || _.chain(Game.rooms).map('controller').flatten().filter('my').map('owner.username').first().value();
   Memory.myRooms = Memory.myRooms || [];
   Memory.squads = Memory.squads || {};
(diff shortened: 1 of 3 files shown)


mschultheiss83 (changes requested):
you should change 
config.nextRoom.maxRooms to be like max(8 (default), Game.gcl.level/2) or something like this

XenoAmess:
> you should change config.nextRoom.maxRooms to be like max(8 (default), Game.gcl.level / 1.5) or something like this

image

What you said is reasonable, and I would use Game.gcl.level to make it sync with existed codes in nextroomers, means `max(config.nextRoom.maxRooms, Game.gcl.level)`

XenoAmess:
@mschultheiss83 please reply and re-review, thanks.

mschultheiss83:
if you have 2 or more nextroomer in 2 different rooms the following line ensures that the first nextroomer wins
`  if (Memory.myRooms.length >= Game.gcl.level) {`

mschultheiss83 (changes requested):
looks good to me, but we should change the config.js
```js
    maxRooms: Math.max(8, Game.gcl.level),
```

XenoAmess:
> looks good to me, but we should change the config.js
> 
> ```js
>     maxRooms: Math.max(8, Game.gcl.level),
> ```

nope as the config will only run once, would never recalculate when you upgrade.

XenoAmess:
> if you have 2 or more nextroomer in 2 different rooms the following line ensures that the first nextroomer wins ` if (Memory.myRooms.length >= Game.gcl.level) {`

yes, it already have such limits. so I don't think we need to add another.

mschultheiss83:
> nope as the config will only run once, would never recalculate when you upgrade gcl level.

we could update it in `brain.prepareMemory`

mschultheiss83:
> yes, it already have such limits. so I don't think we need to add another.
it for after respawn like me GCL 28 but like 6 rooms
so maxRooms should be near you GCL.
i have faced some memory issues when you have more than 20 rooms so i added 
`  config.nextRoom.maxRooms = Math.max(8, Game.gcl.level / 1.5); ` to the `brain.prepareMemory` function

XenoAmess:
> > yes, it already have such limits. so I don't think we need to add another.
> > it for after respawn like me GCL 28 but like 6 rooms
> > so maxRooms should be near you GCL.
> > i have faced some memory issues when i had more than 20 rooms so i added
> > ` config.nextRoom.maxRooms = Math.max(8, Game.gcl.level / 1.5);` to the `brain.prepareMemory` function

@mschultheiss83 I understood what you wanna. please have a look at the latest version, I added a dynamic config stage,  and IMO it can do.

XenoAmess:
@dependabot use this milestone

TooAngel:

PR #664, opened 2022-09-07
[LOGIC REFINE]after claiming, turn claimer to a reserver or a scout, instead of kill it.

XenoAmess:


mschultheiss83 (changes requested):
please check Creep.recycleCreep prototype_creep_startup_tasks.js:101

src/prototype_creep.js
```diff
@@ -102,8 +102,7 @@ Creep.prototype.checkForHandle = function() {
   const role = this.memory.role;
   if (!role) {
     this.log('Creep role not defined for: ' + this.id + ' ' + this.name.split('-')[0].replace(/[0-9]/g, ''));
-    this.memory.killed = true;
-    this.suicide();
+    Creep.recycleCreep(this);
```
mschultheiss83, line 105: i dont know if this is working... did you test it? 
if creep.memory.role is not set can a creep then move to a spawn for recycle?

src/prototype_creep.js
```diff
@@ -117,7 +116,7 @@ Creep.prototype.handle = function() {
   try {
     if (!this.unit()) {
       this.log('Unknown role suiciding');
-      this.suicide();
+      Creep.recycleCreep(this);
```
mschultheiss83, line 119: same as above, and we should change the log message 'Unknown role recycling creep'

src/prototype_creep_clean.js
```diff
@@ -142,8 +142,7 @@ Creep.prototype.cleanSetTargetId = function() {
     }
   }
   this.memory.targetReached = true;
-  this.memory.killed = true;
   this.log('Nothing found, suicide');
```
mschultheiss83, line 145: same here `  this.log('Nothing found, recycling');`

src/prototype_room_controller.js
```diff
@@ -116,5 +116,5 @@ Room.prototype.clearRoom = function() {
   _.each(constructionSites, (cs) => cs.remove());
 
   const creeps = this.findMyCreeps();
-  _.each(creeps, (cs) => cs.suicide());
+  _.each(creeps, (cs) => Creep.recycleCreep(cs));
```
mschultheiss83, line 119: this can not work `_.each(structures, (s) => s.destroy());` so creep will not find a spawn in the room

src/role_carry.js
```diff
@@ -55,15 +55,15 @@ function checkHelperEmptyStorage(creep) {
...
     if (targetStructure === null) {
-      creep.suicide();
+      Creep.recycleCreep(creep);
       return;
     }
 
     if (targetStructure.structureType === STRUCTURE_STORAGE) {
       creep.say('storage');
       if (targetStructure.store.energy === 0) {
         creep.log('Suiciding the storage I should get the energy from is empty');
```
mschultheiss83, line 65: this should work 
but `creep.log('recycling: the storage I should get the energy from is empty');`

src/role_carry.js
```diff
@@ -399,8 +399,7 @@ roles.carry.action = function(creep) {
   // End of path, can't harvest, suicide (otherwise the sourcer gets stuck)
   if (!reverse && creep.body.filter((part) => part.type === WORK).length === 0) {
     // creep.log('Suiciding because end of path, no energy, do not want to get in the way of the sourcer (better recycle?)');
-    creep.memory.killed = true;
-    creep.suicide();
+    Creep.recycleCreep(creep);
```
mschultheiss83, line 402: we should test this

src/role_claimer.js
```diff
@@ -22,7 +22,7 @@ roles.claimer.action = function(creep) {
   const returnCode = creep.claimController(creep.room.controller);
   if (returnCode === OK) {
     creep.creepLog('New claimer, in room, claimed');
-    creep.suicide();
+    Creep.recycleCreep(creep);
```
mschultheiss83, line 25: this should depend on 
```js 
if (creep.memory.timeToTravel > 0 && creep.memory.born - Game.time < creep.memory.timeToTravel) {
// creep will not reach spawn in time
  creep.suicide();
} else {
  Creep.recycleCreep(creep);
}
```

src/role_quester.js
```diff
@@ -16,7 +16,7 @@ roles.quester.settings = {
 roles.quester.questLost = function(creep, quest, reason, value) {
   creep.log(`Quest lost cs: ${value} ${JSON.stringify(quest)}`);
   delete Memory.quests[creep.memory.level];
-  creep.suicide();
+  Creep.recycleCreep(creep);
```
mschultheiss83, line 19: maybe we should add into `Creep.recycleCreep` function 
`creep.memory.timeToTravel > 0 && creep.memory.born - Game.time < creep.memory.timeToTravel` and let the recycleCreep decide between suicide and recycle

src/role_quester.js
```diff
@@ -33,7 +33,7 @@ roles.quester.questWon = function(creep, quest) {
   };
   creep.room.terminal.send(RESOURCE_ENERGY, 100, quest.player.room, JSON.stringify(response));
   delete Memory.quests[creep.memory.level];
-  creep.suicide();
+  Creep.recycleCreep(creep);
```
mschultheiss83, line 36: same here timeToTravel vs born

mschultheiss83 (changes requested):
i guess patching the Creep.recycleCreep function would be necessary
`creep.memory.timeToTravel > 0 && creep.memory.born - Game.time < creep.memory.timeToTravel` 
and let the recycleCreep decide between suicide and recycle

src/prototype_creep.js
XenoAmess: Good question.
Before, I use the new added implemented function, so I know it can, but now it is an existed function...
I will give it a role here, so it would think itself be a scout.

src/role_claimer.js
XenoAmess: I will refine the existed recycleCreep function, and the check would be added there, using getRoomLinearDistance.

src/role_quester.js
XenoAmess: > maybe we should add into `Creep.recycleCreep` function `creep.memory.timeToTravel > 0 && creep.memory.born - Game.time < creep.memory.timeToTravel` and let the recycleCreep decide between suicide and recycle

yes! that is what I mean

src/prototype_room_controller.js
XenoAmess: now it will search for other room.

src/role_carry.js
XenoAmess: yes, this is what I worried about.
In worst cases might cause lots of carriers stuck together

src/role_quester.js
XenoAmess: added the judge into recycleCreep function in

PR #668, opened 2022-09-08
[SEVERE]fix a bug in getPriority about room name, when targetRoom be an object

XenoAmess:


mschultheiss83 (commented):
when did this happen?

Issue #674, opened 2022-09-10
[LOGIC REFINE] suggest spawn creeps at TOP, but not force it.


XenoAmess:
is there any reason to force the position we spawn creeps at top?
I do wonder because sometimes when I get severe stuck, I think maybe it be better for sometime allow it spawn from other sides
any ideas?

XenoAmess:
Well I did a change commit about it and during one whole day in 4 rooms, I don't see it make any trouble. think should pr it here.

XenoAmess:
also agreed by @mschultheiss83 in discord, so I think we can try it at least..

mschultheiss83:
sure, let's test

TooAngel:
The planing of the base puts spawns below (bottom) of the path. That is the reason that it always spawns to the top.
Without this it can happen, that the creep will be on a position which can not reach the main path, because the extensions are blocking the way.

PR #675, opened 2022-09-10
[WIP][LOGIC REFINE] allow spawn creep to other than top

XenoAmess:
close https://github.com/TooAngel/screeps/issues/674


mschultheiss83:
btw, you should check for the main-path and maybe just allow to spawn creeps on the street... it else happen that our ceeps get stuck behind our structures, right?

TooAngel:

PR #677, opened 2022-09-11
[BUG]fix bug in construction target finding in role squadsiege.

XenoAmess:
now it will not try to dismental controller anymore.
besides `this.pos.findClosestByPath(FIND_HOSTILE_STRUCTURES, STRUCTURE_TOWER);` will actually search for any structure, with wrong use of `opt` param, so fix it.


TooAngel:

PR #681, opened 2022-11-03
[LOGIC REFINE]prevent sending useless reservers

TSStudio:
According to https://docs.screeps.com/start-areas.html
> The sector is completely separated from the outer world with a blind, indestructible wall preventing outside players from entering it. Getting inside a Novice Area is possible only by placing there your initial spawn directly.

It's impossible to travel between rooms with different status, so sending reservers is just wasting energy.  
Now check before sending them.

Updated Nov 6, 2022:

In a reserved room in a novice area, and next to normal area, walls exist but cannot be destroyed. Reservers shouldn't call structurers if the structure is not able to be destroyed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4427,9 +4427,9 @@
       "dev": true
     },
     "loader-utils": {
-      "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/loader-utils/-/loader-utils-1.4.0.tgz",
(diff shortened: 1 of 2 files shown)


mschultheiss83 (approved):
Nice thank you, looks good to me

mschultheiss83:
image
same goes for scouts

TSStudio:
> image
> same goes for scouts

Working on it, should finish in a few days. By the way, I’d like to know whether only scouts, reservers and structurers go to nearby rooms? Am I missing any other roles?

TSStudio:
Now it should work for all roles.

mschultheiss83:
can you rebase, please?

TSStudio:
Yes sure, but there's some mess when I rebase, hopefully merging won't be a big problem.

TooAngel:

PR #688, opened 2023-01-08
Improve setup documentation and increase version

TooAngel:
Mostly increasing the version so that it will be released.


TooAngel:

PR #689, opened 2023-01-23
Fix cli on test server

TooAngel:
Due to the latest docker setup changes the CLI port of the screeps server also needs to be exposed on 0.0.0.0.


TooAngel:

PR #690, opened 2023-01-30
Cleaning up eslint configs

TooAngel:
Merging all config into the `package.json`. I don't think anyone kept track of the different configs, also it clutters the root of the directory.


TooAngel:

PR #693, opened 2023-02-20
Refactor brain

TooAngel:
Changing methods from binding them to the global `brain` object to standard functions and exports / require the ones which are used in other modules. This makes it easier to understand where which function is used.


mschultheiss83 (approved):
If I get it right, this is just cosmetics. So approved 😊

TooAngel:

TooAngel:
Yeah, mostly

PR #695, opened 2023-02-25
bugfix: don't try to produce cliamer in a room that cannot.

XenoAmess:


src/prototype_room_creepbuilder.js
```diff
@@ -473,7 +475,7 @@ Room.prototype.getCreepConfig = function(creep) {
 Room.prototype.spawnCreateCreep = function(creep) {
   const spawns = this.findSpawnsNotSpawning();
   if (spawns.length === 0) {
-    return;
+    return false;
```
mschultheiss83, line 478: please change @return {boolean|void} to @return {boolean} in line 473 also

src/prototype_room_creepbuilder.js
```diff
@@ -54,6 +54,8 @@ Room.prototype.spawnCheckForCreate = function() {
   creep.ttl = creep.ttl || config.creep.queueTtl;
   if (this.findSpawnsNotSpawning().length === 0) {
     creep.ttl--;
+  } else if (this.energyAvailable === this.energyCapacityAvailable) {
```
mschultheiss83, line 57: please look into  `.checkRoleToSpawn('claimer'`

mschultheiss83 (changes requested):

src/prototype_room_creepbuilder.js
TooAngel: Nah, this would clear the spawning queue, when all extensions are filled and all spawns are currently spawning.

TooAngel (changes requested):
I doubt this would work out. If there is a bug with the claimer (e.g. it tries to spawns but the rooms does not have enough max energy). The claimer shouldn't be queued at all.

PR #696, opened 2023-02-25
feature: add allowWall and allowFlag mechanism

XenoAmess:


mschultheiss83:
why do you need this?

TooAngel (commented):
Hm, I'm fine with this. A bit more explanation in the commit or in the documentation would be nice, so that other know about it.

XenoAmess:
For building some walls on the place which I humanly think it should.

If not building such a mechanism, bot will destroy the walls you clicked.

Could be used to build some wall-arts, or for some extream defensive point that can't be found by bot.

TooAngel (approved):
Will approved this now.

But please try to be a bit more detailed in the commit message and pull request description.
In a couple of weeks it's tricky to remember what this was good for.

PR #697, opened 2023-02-25
feature: should treat Source Keeper samely like Invader, as they are both malicious npc.

XenoAmess:


src/main.js
```diff
@@ -23,6 +23,7 @@ global.data.stats = global.data.stats || {
   memoryFree: 0,
   heapFree: 0,
 };
+global.maliciousNpcUsernames = ['Invader', 'Source Keeper'];
```
TooAngel, line 26: Please move this into the config instead.

TooAngel (approved):

src/main.js
XenoAmess: @TooAngel done.

TooAngel (approved):
I think you can access the config without `global.`, too - so just `config.maliciousNpcUsernames`. But it's fine.

Good job.

PR #700, opened 2023-02-25
refine: refine sourcer pickup energy logic. Also allow build container in base, when have no link nearby.

XenoAmess:
…

### DIFF
--- a/src/role_sourcer.js
+++ b/src/role_sourcer.js
@@ -183,14 +183,7 @@ function getContainerConstructionSite(creep) {
  */
 function maintainContainer(creep) {
   if (creep.inBase()) {
-    const links = creep.pos.findInRangePropertyFilter(FIND_STRUCTURES, 1, 'structureType', [STRUCTURE_LINK]);
-    if (links.length) {
-      const containerConstructionSite = getContainerConstructionSite(creep);
-      if (containerConstructionSite) {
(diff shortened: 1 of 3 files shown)


src/role_sourcer.js
```diff
@@ -183,7 +183,14 @@ function getContainerConstructionSite(creep) {
...
   if (creep.inBase()) {
-    return false;
+    const links = creep.pos.findInRangePropertyFilter(FIND_STRUCTURES, 1, 'structureType', [STRUCTURE_LINK]);
+    if (links.length) {
+      const containerConstructionSite = getContainerConstructionSite(creep);
+      if (containerConstructionSite) {
+        containerConstructionSite.remove();
+      }
+      return false;
+    }
```
TooAngel, line 193:
```suggestion
    return creep.room.controller.level < 6;
```

How about this?
Doing find operations per tick per sourcer costs CPU.
At RCL 6 we should have all links at the sourcer.
If we don't maintain the container they will decay at some point in time, so no need to touch.
What do you think?

TooAngel (changes requested):

src/role_sourcer.js
XenoAmess: sounds good!

TooAngel (approved):

TooAngel:

PR #701, opened 2023-02-25
feature: allow up to 2 upgraders at same time.

XenoAmess:


mschultheiss83 (commented):
I do not like this

TooAngel (commented):
Can you please explain the case a bit more in detail?

I have the impression that we do not have too much energy usually.
Currently a upgrader is build the way that it will only take half of the current storage. If you want to speed this up / optimize on controller upgrade. I would suggest to update the upgrader spawn config (bigger upgraders), put the factor into the config so that it can be overwritten.

This changes here would only help in case we are always building the max upgrader.

PR #702, opened 2023-02-25
feature: don't build too crowded around controller.

XenoAmess:


mschultheiss83 (commented):
you did this cause of https://github.com/TooAngel/screeps/pull/701
if so we don't need this

XenoAmess:
> you did this cause of #701 if so we don't need this

mostly, but also for the grand crowd around the controller

TooAngel (commented):
Not sure about this one. I did it on purpose somehow small to be able to handle also tiny rooms.

TooAngel:

PR #703, opened 2023-02-25
fixed: mineral handling

mschultheiss83:
moved handleReaction to prototype_room_mineral.js
added Room.prototype.getEnergy

### DIFF
--- a/src/prototype_room_mineral.js
+++ b/src/prototype_room_mineral.js
@@ -221,3 +221,78 @@ Room.prototype.handleTerminal = function() {
   delete this.memory.mineralOrder[roomOtherName];
   return true;
 };
+
+Room.prototype.handleReaction = function() {
+  if (!this.memory.reaction) {
+    return false;
+  }
+
+  let labsToReact = this.findStructuresOfStructureType(STRUCTURE_LAB).filter((lab) => {
+    if (lab.store[this.memory.reaction.result.first] > 0 || lab.store[this.memory.reaction.result.second] > 0) {
+      return false;
+    }
+    return true;
+  });
+
+  if (labsToReact.length === 0) {
+    return false;
+  }
+
+  const mineralCreep = this.findMyCreepsOfRole('mineral').length ? this.findMyCreepsOfRole('mineral')[0] : false;
+  const cleanupNotEnoughResources = () => {
+    if (mineralCreep && this.memory.reaction && this.memory.reaction.result) {
+      if (!mineralCreep.checkLabEnoughMineral(lab1, this.memory.reaction.result.first) || !mineralCreep.checkLabEnoughMineral(lab2, this.memory.reaction.result.second)) {
+        roles.mineral.cleanUpLabs(mineralCreep);
+      }
+    }
+  };
+
+  const lab0 = Game.getObjectById(this.memory.reaction.labs[0]);
+  const lab1 = Game.getObjectById(this.memory.reaction.labs[1]);
+  const lab2 = Game.getObjectById(this.memory.reaction.labs[2]);
+  if (lab0 === null || lab1 === null || lab2 === null) {
+    delete this.memory.reaction;
+  } else {
+    const labsToReactResponse = labsToReact.map((lab0) => {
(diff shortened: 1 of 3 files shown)


src/prototype_room_mineral.js
```diff
@@ -221,3 +221,87 @@ Room.prototype.handleTerminal = function() {
...
   return true;
 };
+
+Room.prototype.cleanupNotEnoughResources = function() {
+  if (!this.memory.reaction) {
+    return false;
+  }
+  const lab1 = Game.getObjectById(this.memory.reaction.labs[1]);
+  const lab2 = Game.getObjectById(this.memory.reaction.labs[2]);
+  const mineralCreep = this.findMyCreepsOfRole('mineral').length ? this.findMyCreepsOfRole('mineral')[0] : false;
```
TooAngel, line 231:
```suggestion
  const mineralCreeps = this.findMyCreepsOfRole('mineral');
  if (mineralCreeps.length === 0) {
      return;
  }
  const mineralCreep = mineralCreeps[0];
```

(minimize find calls, to save CPU)

src/prototype_room_mineral.js
```diff
@@ -221,3 +221,87 @@ Room.prototype.handleTerminal = function() {
...
+  if (mineralCreep && this.memory.reaction && this.memory.reaction.result) {
```
TooAngel, line 232:
```suggestion
  if (this.memory.reaction.result) {
```

`this.memory.reaction` is already checked at the beginning of the method.

src/prototype_room_mineral.js
```diff
@@ -221,3 +221,87 @@ Room.prototype.handleTerminal = function() {
...
+
+  if (labsToReact.length === 0) {
+    return false;
+  }
+
+  const lab0 = Game.getObjectById(this.memory.reaction.labs[0]);
+  const lab1 = Game.getObjectById(this.memory.reaction.labs[1]);
+  const lab2 = Game.getObjectById(this.memory.reaction.labs[2]);
+  if (lab0 === null || lab1 === null || lab2 === null) {
+    delete this.memory.reaction;
```
TooAngel, line 260:
```suggestion
    delete this.memory.reaction;
    return;
}
```

 the code is a bit cleaner

src/prototype_room_my.js
```diff
@@ -357,8 +357,28 @@ Room.prototype.handleReputation = function() {
...
+  if (this.storage) {
+    amount += this.storage.store[RESOURCE_ENERGY];
+  }
+  return amount;
+};
+
 Room.prototype.getBuilderAmount = function(constructionSites) {
   let amount = 1;
+  const energyInRoom = this.getEnergy();
+  if (energyInRoom > (config.terminal.minEnergyAmount / 2)) {
```
TooAngel, line 379: :-) What does this mean?

TooAngel (commented):
:tada: I like to see, that there is progress on mineral handling and boosting.

Left some smaller comments. I need a bit more time to review and understand it fully.

src/prototype_room_my.js
mschultheiss83: there should be more Builder if we have energy in terminal and are upgrating to RCL 6/7/8, there are a lot of buildings that needs to be done asap (e.g links near 2nd sourcer)

mschultheiss83:
basically thats the first part of mineral getting back to work

src/prototype_room_my.js
TooAngel: This is tricky to understand from that line.
Wouldn't it be better to use `this.isHealthy()`? And maybe extend this. This is reusable and makes clearer what the intention is.

src/prototype_room_mineral.js
```diff
@@ -221,3 +221,89 @@ Room.prototype.handleTerminal = function() {
...
+    return false;
+  } else {
```
TooAngel, line 263:
```suggestion
  }
```
(move the `}` up)
The `if` clause ends with a `return`. It is a bit easier to read.

TooAngel (changes requested):
The builder logic should be improved. This rest is hopefully fine :-)

PR #704, opened 2023-02-25
More brain refactoring and reducing max-statements

TooAngel:
Another round of cleaning up code.
Converted some more function in brain to actual functions. Improved some methods with high max-statements.


TooAngel:

PR #705, opened 2023-02-26
Fixing last merged issues

TooAngel:
- The spawn direction can only by TOP right now, as the room layout is designed that way.
- When starting the private server `getRoomStatus` is not directly available => Catching the exception
- `base` is not always provided when spawning a creep, using the room name in that case.


TooAngel:

PR #706, opened 2023-02-26
PIXEL selling and buyEnergy

mschultheiss83:
- cleanup pixel code, sell PIXELs
- buyEnergy if we have terminals and storage, mineral creeps should move energy to storage for builders and upgrader (see PR https://github.com/TooAngel/screeps/pull/703)
- fixed quests.js for incoming transactions

### DIFF
--- a/src/brain_pixel.js
+++ b/src/brain_pixel.js
@@ -1,0 +1,40 @@
+'use strict';
+
+const {debugLog} = require('./logging');
+const pixelLog = (message) => {
+  debugLog('pixel', message);
+}
+
+module.exports.handlePixel = function() {
+  if (typeof PIXEL !== 'undefined') {
+    // generate PIXEL
+    if (Game.cpu.bucket >= PIXEL_CPU_COST + config.pixel.minBucketAfter) {
+      global.load = Math.round(Game.cpu.getUsed());
+      Game.cpu.generatePixel();
+      const pixel = Game.resources[PIXEL] || ""
+      const deltaPixel = Game.time - (global.data.pixel || 0)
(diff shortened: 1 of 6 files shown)


src/brain_pixel.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
```
TooAngel, line 1: I'm not sure about the `brain` prefix anymore. I would just call it `pixel.js`

src/config.js
```diff
@@ -289,11 +292,15 @@ global.config = {
...
+    buyEnergy: {
+      enabled: true,
+      allowedSalesHistoryDeviation: 0.05,
+    },
     // buy power if we have more credits than config.market.minCredits
     buyPower: false,
-    // 300M credits
-    minCredits: 300000000,
+    // 300K credits
+    minCredits: 300000,
```
TooAngel, line 303: I think we should have as minCredits the price of an access key, which is currently around 500M.
So we shouldn't reduce this, if you prefer you can change it in your local config.

What do you think?

src/quests.js
```diff
@@ -129,17 +126,22 @@ function getQuestFromTransactionDescription(description) {
  * checkQuestForAcceptance
  *
  * @param {object} transaction
- * @return {bool}
+ * @return {boolean}
  */
 function checkQuestForAcceptance(transaction) {
   Memory.quests = Memory.quests || {};
+  transaction.description = transaction.description || JSON.stringify({
```
TooAngel, line 133: I don't understand this change.
This sets the description of the transaction to some 'random' value.

TooAngel (changes requested):
Looks good.

Only the config changes and the default values on transcations needs to be improved or discussed.

PR #707, opened 2023-02-26
small fixes and improvements

mschultheiss83:
- respawnMe

improvement for room visuals:
- added `1st Role/Queue`

fixes:
- handleReviveRoomQueueCarry: controller level should be at level 2 or above
- isConstructingSpawnEmergency: wrong order => set value before reading it
- checkForRoute: should use highway
- reviveMyNow: do not send nextroomer if there are hostiles => less energy waste
- some spaces
- typo/missing spaces in prototype_room_find.js
- harvestCommodities: move minRCL and range to config
- findAttackCreeps: do not attack friends
- ttl in creep queue, higher controller level => shorter ttl
- typo/missing spaces in prototype_room_costmatrix.js
- useHighway more often for carry, claimer and nextroomer
- avoid hostile rooms
- repairer found energy behind a wall, can not pickup
- new line before: module.exports
- hostileCreepCount not found
- suppress Deleting from memory for empty objects

### DIFF
--- a/src/brain_memory.js
+++ b/src/brain_memory.js
@@ -58,11 +58,11 @@ function handleUnexpectedDeadCreeps(name, creepMemory) {
     console.log(`${Game.time} ${name} handleUnexpectedDeadCreeps no creepMemory.room creepMemory: ${JSON.stringify(creepMemory)} data: ${JSON.stringify(global.data.creeps[name])}`);
   }
 
-  debugLog(`${Game.time} ${creepMemory.room} ${name} memory hostile: ${data.hostileCreepCount}`);
   if (data.hostileCreepCount > 0) {
+    debugLog(`${Game.time} ${creepMemory.room} ${name} memory hostile: ${data.hostileCreepCount}`);
     debugLog('brain', `${creepMemory.room} ${name} Not in Game.creeps with hostiles lived ${Game.time - creepMemory.born} hostiles: ${data.hostileCreepCount} - I guess killed by hostile`); // eslint-disable-line max-len
   } else {
-    console.log(`${Game.time} ${creepMemory.room} ${name} Not in Game.creeps without hostiles lived ${Game.time - creepMemory.born} hostiles: ${data.hostileCreepCount}`); // eslint-disable-line max-len
+    console.log(`${Game.time} ${creepMemory.room} ${name} Not in Game.creeps without hostiles lived ${Game.time - creepMemory.born}`); // eslint-disable-line max-len
   }
 
   if (Game.time - creepMemory.born < 20) {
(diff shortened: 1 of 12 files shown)


src/brain_memory.js
```diff
@@ -169,7 +169,9 @@ function cleanRooms() {
         delete Memory.rooms[name];
       }
       if (Memory.myRooms.indexOf(name) < 0) {
-        console.log(Game.time, `Deleting ${name} from memory, no myRoom ${JSON.stringify(Memory.rooms[name])}`);
+        if (Object.keys(Memory.rooms[name]).length > 0) {
```
TooAngel, line 172: So, this idea is to only store `myRooms` in Memory. All other rooms can be stored in the heap and recalculated. This makes sure that we don't reach the memory limit.

Currently some other rooms are also stored in memory, but mostly with empty values.

Do you know a case where we want to have external rooms stored in memory?

src/prototype_room_creepbuilder.js
```diff
@@ -51,7 +51,7 @@ Room.prototype.spawnCheckForCreate = function() {
     this.memory.queue.shift();
     return false;
   }
-  creep.ttl = creep.ttl || config.creep.queueTtl;
+  creep.ttl = creep.ttl || Math.ceil(config.creep.queueTtl * _.size(CONTROLLER_DOWNGRADE) / this.controller.level);
```
TooAngel, line 54: Can you please add a comment or extract it in a method with proper naming, so that it is clearer what it does.

I have no clue why to integrate downgrade or controller level.

src/prototype_room_creepbuilder.js
```diff
@@ -195,9 +195,11 @@ Room.prototype.checkRoleToSpawn = function(role, amount, targetId, targetRoom, l
   if (this.inQueue(creepMemory) || this.inRoom(creepMemory, amount)) {
     return false;
   }
-  if (targetRoom !== base && typeof Game.map.getRoomStatus(targetRoom) !== 'undefined' && typeof Game.map.getRoomStatus(base) !== 'undefined') {
-    if (Game.map.getRoomStatus(targetRoom).status !== Game.map.getRoomStatus(base).status) {
-      return false;
+  if (targetRoom !== base) { // should fix ci Error: No runtime status data
+    if (Game.time > 230 && typeof Game.map.getRoomStatus(targetRoom) !== 'undefined' && typeof Game.map.getRoomStatus(base) !== 'undefined') {
```
TooAngel, line 199: I added a try catch block.

But `Game.map.getRoomStatus` should only be called once and stored as variable, to save CPU and looks cleaner.

src/prototype_creep_routing.js
```diff
@@ -12,8 +12,16 @@ Creep.prototype.getRoute = function() {
   }
 
   let route = [];
+  let useHighWay = false;
+  switch (this.memory.role) {
+  case 'carry':
+  case 'claimer':
+  case 'nextroomer':
+    useHighWay = true;
+  }
```
TooAngel, line 21: Not sure about this.

Other creeps like `sourcer` or `reserver` might need to route through highway rooms as well. So which creeps shouldn't move throw highway rooms? `repairer` and `universal`, ... - creeps which shouldn't route at all.

TooAngel (changes requested):
Left some comments.

PR #708, opened 2023-02-27
Cleanup the user utils

TooAngel:
The `src/utils.js` module should provide function to the user, to be used via the console.


TooAngel:

PR #709, opened 2023-03-04
Refactor: Remove findClosestByRangePropertyFilter

TooAngel:
Instead of abstracting the screeps API, we should extract as named functions which can be reused.

This allows to easier read the code and also allows to reused certain functions, as well as changes these at a common place in case of updates.

This is step one for removing all propertyfilter methods.


TooAngel:

PR #711, opened 2023-03-04
Introduce `spawnIdle`

TooAngel:
Store for each of `myRooms` the time a spawn is idle. Include this in the reserver logic and restrict new reservation on a spawnIdle threshold.

Currently too many rooms are tried to be reservered, while the actual `reserver` creeps are queue due to busy spawns.
The `spawnIdle` check tackles this issue.


TooAngel:

PR #716, opened 2023-11-05
reduce noise

TooAngel:
Disable commodities and attack log by default.


TooAngel:

